PR #3, opened 2015-05-03
[refactor] nicer image (VOLUME/style) + docs

thibaultCha:
#### Changes on the Hub:

As already discussed internally, the current image on the hub is named wrong ([mashape/docker-kong](https://registry.hub.docker.com/u/mashape/docker-kong/)).

Since there is no way to rename an image, I created [mashape/docker](https://registry.hub.docker.com/u/mashape/kong/).

Instead of having to use the image as:

```shell
$ docker pull mashape/docker-kong
```

This one will would feel nicer:

```shell
$ docker pull mashape/kong
```

**Maybe give a thumb up if you want to drop mashape/docker-kong in favour of mashape/docker on the hub.**

#### Changes in this PR

This PR also proposes a rewrite of the usage of this image, as well as a slight refactor of the said image:

- It uses a [Volume](https://docs.docker.com/userguide/dockervolumes/) for the configuration
- It is more readable (as advised by docker's documentation)

We should also rename the [mashape/docker-cassandra](https://github.com/spotify/docker-cassandra) image to mashape/cassandra (and give it a volume too)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,13 +7,13 @@ RUN yum -y install dnsmasq wget
 # configuring dnsmasq
 RUN echo -e "user=root\nno-resolv\nserver=8.8.8.8" >> /etc/dnsmasq.conf
 
-# download Kong
-RUN wget https://github.com/Mashape/kong/releases/download/0.2.0-2/kong-0.2.0_2.el7.noarch.rpm && yum install -y kong-0.2.0_2.el7.noarch.rpm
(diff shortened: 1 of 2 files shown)


ahmadnassri:
:+1: much better.

thibaultCha:
Also we need to comply to these: https://docs.docker.com/docker-hub/official_repos/ if we want Kong to make it as an official image.

sinzone:
just "Kong" is the ideal one.

thibaultCha:
If official, then yes, first is needs to be mashape/kong

PR #11, opened 2015-08-07
Update Dockerfile to always get latest

ahmadnassri:


thefosk:
We tag our Kong Dockerfiles so we can provide all the version, including `latest`. This change would prevent us from tagging versions, because it would always get the latest one no matter what.

sinzone:
also, tagging dockerfiles is part of the process for becoming official docker image.

ahmadnassri:
there are two types of docker tagging, you can have a "latest" tag that's always up to date, and a versioned one.

with downloadkong.org, you can add `?version=x.x.x` and get the specific version

montanaflynn:
Hey @thefosk and @ahmadnassri, I hope you are both doing well :smiley: 

1) https://github.com/Mashape/docker-kong/blob/latest/Dockerfile is pointing to an old version of Kong, this should be updated to either use downloadkong.org or be updated as part of the release distribution script.

2) https://hub.docker.com/r/mashape/kong/~/dockerfile/ is similarly pointing to an old version of Kong. I don't know why it's not pointing to the Dockerfile in the master branch which is in fact the latest version.

ahmadnassri:
hey @montanaflynn :wave: 

yes, seems weird that `hub.docker.com` is showing the very first version, but running locally runs the correct current version `0.4.1`, I'm guessing the website must be a bug or something else on their end ...

regardless, we should address our "latest" tagging approach, currently its all manual and `downloadkong.org` only solves 50% of the problem

thibaultCha:
At the end of the day we still need a version number in it for readability/maintainability (people can easily change it)

So no point in going through another service which can be another point of failure. GitHub releases will do just fine.

thefosk:
I share @thibaultCha 's point of view. Since versions should be manually specified anyways, then we don't need an additional redirect service.

thefosk:
@montanaflynn @ahmadnassri I noticed the Docker Hub problem with the Dockerfile, and I think it's a bug on their side. They should always get the `latest` Dockerfile. 

Regarding https://github.com/Mashape/docker-kong/blob/latest/Dockerfile, do we need it? We have crated a tag called `latest`, but we don't really need it.

ahmadnassri:
yep. I'm coming around to @thibaultcha's point.

also, certainly a Docker hub bug.

thefosk:
Closing this then.

montanaflynn:
About the outdated Dockerfile showing on `hub.docker.com` here's their response:

> It looks like there are several tags that get built. Each one could theoretically have its own Dockerfile. Unfortunately we only show the Dockerfile for the last build to finish.

ahmadnassri:
that's weird ... so do they advice to what is a best practice here in
tagging / versioning?

-- 


---
Ahmad Nassri

montanaflynn:
They allow for two methods of tagging images with automated builds (which is what is applicable in this case) including: 

a) Branches
b) Git Tags

Both of these options also allow for using paths. Many open source projects don't use git tags and only have a master branch and rely on creating seperate directories to handle versions.

thefosk:
We have a `latest` tag on DockerHub that points to a specific tag to our repository. Since `latest` is their reserved word for the latest version, I don't understand why they don't just use that.

montanaflynn:
> regardless, we should address our "latest" tagging approach, currently its all manual and downloadkong.org only solves 50% of the problem

@ahmadnassri what's the problem with what we're doing currently? The only thing I can think of is that it could be automated as part of the release distribution scripts. Like @thefosk said, the `latest` tag is 
defined from within `hub.docker.com` where there are both tagged older versions along with a latest tag that is working just fine.

> Since latest is their reserved word for the latest version, I don't understand why they don't just use that.

@thefosk That's exactly what I suggested to them along with the ability to see all the tags and associated Dockerfiles from that section. They are working on it :+1:

thefosk:
> They are working on it 

@montanaflynn :+1:

PR #48, opened 2016-08-17
Fix proposal for issue #1074

orphaner:
There are a lot of serf defunct process while running kong in a docker
container.
A docker container is meant to run only one process. But the kong start
command runs serf, dnsmasq and many other processes. All these process pid's
don't have a parent and they become a zombie process when they are
finished.
The fix for this is to use a lite init system such as tini, or dumb-init.

Here is a fix proposal using tini.


thefosk:
Already fixed.

PR #63, opened 2016-12-31
Uses docker overlay

hutchic:
- use docker overlay for service discovery / routing instead of consul / containerpilot
- upgrade docker-compose file to docker-compose version 2
- ~had to add a wait_for_postgres.sh to the kong container~ no longer true with docker-compose 2.1


hutchic:
The following demonstrates why the wait_for_postgres needed to be added

```
vagrant@vagrant:docker|master⚡ ⇒  docker-compose -f docker-compose.kong.yml up -d kong
Creating kong-postgres
Creating docker_kong_1

vagrant@vagrant:docker|master⚡ ⇒  docker logs docker_kong_1                           
Error: /usr/local/share/lua/5.1/kong/cmd/start.lua:18: [postgres error] connection refused

vagrant@vagrant:docker|master⚡ ⇒  docker-compose -f docker-compose.kong.yml kill kong 
Killing docker_kong_1 ... done

vagrant@vagrant:docker|master⚡ ⇒  docker-compose -f docker-compose.kong.yml rm kong
Going to remove docker_kong_1
Are you sure? [yN] y
Removing docker_kong_1 ... done

vagrant@vagrant:docker|master⚡ ⇒  docker-compose -f docker-compose.kong.yml up -d kong
kong-postgres is up-to-date
Creating docker_kong_1

vagrant@vagrant:docker|master⚡ ⇒  docker logs docker_kong_1                           
vagrant@vagrant:docker|master⚡ ⇒  
```

thefosk:
We should proxy using `stream` directives, so we can transparently proxy TLS connections too on ports `8443` and `8444` and let Kong handle all the certificates. @hutchic

hutchic:
I think with everything as it currently is it works would make a good spot to merge. We can migrate to `stream` in a separate PR. Also the `docker-compose` should really be considered just a proof of concept as it won't handle SSL at all right now

thefosk:
The current integration already supports SSL ports, we shouldn't remove support for existing features.

hutchic:
The current docker-compose doesn't support SSL so this wouldn't be a regression

thefosk:
@hutchic SSL support is implemented here https://github.com/Mashape/docker-kong/blob/master/compose/nginx/nginx.conf.ctmpl#L27-L39

You can try yourself by starting up the compose, and consuming `https://127.0.0.1:8443/` or `https://127.0.0.1:8444/`.

hutchic:
Right but what certificate is that going to use ie this compose file is really a good example but I don't think it's ever been / should be considered production ready.

Maybe I'm mistaken though

thefosk:
The certificate being used it's the one that every Kong node uses. The load balancer server it's defined in a NGINX's `stream` entry, which is basically a TCP load-balancer.

We don't have to define any cert in NGINX because whatever incoming request is being directly load-balanced to the underlying Kong node, including TLS connections.

This PR should also implement load-balancing in a `stream` directive so that we can let the underlying Kong node deal with TLS.

hutchic:
ah ok so upstream is doing SSL cool I'll adjust 👍

hutchic:
@thefosk I did the change although determined later it's not necessary / doesn't improve anything. There's no difference in 
- http://nginx.org/en/docs/stream/ngx_stream_core_module.html vs
-  http://nginx.org/en/docs/stream/ngx_stream_proxy_module.html

additionally the current setup doesn't support SSL that I can see
```
http POST localhost:8001/apis/ name=exampole-api hosts=example.com upstream_url=https://httpbin.org

http GET https://localhost:8443 Host:example.com

http: error: SSLError: HTTPSConnectionPool(host='localhost', port=8443): Max retries exceeded with url: / (Caused by SSLError(SSLError("bad handshake: Error([('SSL routines', 'ssl3_get_server_certificate', 'certificate verify failed')],)",),)) while doi
ng GET request to URL: https://localhost:8443/
```

I feel strongly that we should merge this as is and subsequent improvements can go in a new PR:
- removes consul as a dependency
- removes servicecontainer as a dependency
- uses the dockeroverlay network
- upgrades the docker-compose version
- has the same features / functionality of the previous solution

thefosk:
@hutchic in my tests the stream module fixes TLS because it delegates to Kong the certificate handshake. Can you explain what behavior are you experiencing?

hutchic:
@thefosk exactly what I posted

on master of docker-kong in the compose directory
```
docker-compose up -d

http POST localhost:8001/apis/ name=exampole-api hosts=example.com upstream_url=https://htt
pbin.org

http GET http://localhost:8000 Host:example.com #works

http GET https://localhost:8443 Host:example.com #doesnt

```

can you post how you get SSL/TLS to work through Kong?

PR #72, opened 2017-03-10
This will avoid the permission error when launching a container

smashingx1:
This is the error I'm trying to fix:

"container_linux.go:247: starting container process caused "exec: \"/docker-entrypoint.sh\": permission denied"


thefosk:
This error should really never happen. Can you try with latest version of the `kong` image? The permissions are already `+x`.

salazar:
Closing this; both `centos` and `alpine` entrypoints have the `x` bit set.

PR #90, opened 2017-07-06
Add Docker Healthcheck

JnMik:
Add Docker Healthcheck to prevent connexion refused when scaling up Kong containers in a Docker swarm service.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,5 +12,11 @@ RUN wget -O /usr/local/bin/dumb-init https://github.com/Yelp/dumb-init/releases/
 COPY docker-entrypoint.sh /docker-entrypoint.sh
 ENTRYPOINT ["/docker-entrypoint.sh"]
 
+# The first kong container will have to run migrations on the database, so it will
+# take a while for the port 8000 to be reachable. --start-period will help us delay
+# the healthcheck failure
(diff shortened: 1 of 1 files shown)


JnMik:
Just realized --start-period is only available on Docker 17.05 tho..  So how should we handle this ?

hutchic:
for now leave out `--start-period` the healthchecks prior to 17.05 just loiter in `starting` for a bit longer then ideal which is fine IMO

JnMik:
If I remove --start-period, healthcheck of the first Kong container will fail plenty of times before it gets ready. So swarm will kill the container before he finish running the migrations.

hutchic:
You sure? Earlier versions of Docker / Docker swarm what I would see in services that take time to start is the healthcheck would idle at `healthcheck: starting`

JnMik:
Well it is what I was experiencing yesterday with Docker 17.05.0-ce, build 89658be without the --start-period option. I don't exactly at which moment healthcheck exit "starting" mode and start counting failures. What triggers the change of state ?

JnMik:
Base on what endeepak

JnMik:
Miss-clicked!  Based on what Endeepak mentionned here https://github.com/Mashape/docker-kong/issues/89 I guess I could drop the pull request.

Anyone have interest that I keep it open ?

PR #99, opened 2017-08-22
fix: use postgres user for kong-database healthcheck

thibaultcha:
This following recurring log was a hint that some recurrent task would not use
an existing PostgreSQL user:

    kong-database_1   | FATAL:  role "root" does not exist

The proposed fix ensures `healthcheck` is always using the default
`postgres` user instead of `$(whoami)`.

We switch `CMD-SHELL` to `CMD` so we can pass arguments to `pg_isready`
in subsequent elements of the Array.

It seems the shell does not inherit the environment variables passed to
the `environment` attribute, so using `$POSTGRES_USER` did not work in
my initial tests.

### DIFF
--- a/compose/docker-compose.yml
+++ b/compose/docker-compose.yml
@@ -49,7 +49,7 @@ services:
       - POSTGRES_USER=kong
       - POSTGRES_DB=kong
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready"]
+      test: ["CMD", "pg_isready", "-U", "postgres"]
       interval: 10s
       timeout: 5s
       retries: 5


thibaultcha:
Additionally, it seems the Kong still starts even though the database container isn't healthy (the command errors out) and we specified:
```yaml
depends_on:
  kong-database:
    condition: service_healthy
```

It seems docker-compose's `depends_on` feature has a few issues online, but I did not investigate and do not know where the issue comes from.

marcoam:
Thanks for writing this @thibaultcha!

thibaultcha:
@marcoam Thanks for the report ;)

p0pr0ck5:
This seems sane to me. @salazar @hutchic ?

PR #100, opened 2017-08-22
feat(alpine) add checksum check

salazar:
Add checksum check to file downloaded in Alpine Dockerfile.


shashiranjan84:
@salazar  LGTM.

PR #104, opened 2017-08-31
fix(compose) upgrading to 9.5

thefosk:


thefosk:
Seems like this PR is actually not required for CE editions of Kong, if @p0pr0ck5 confirms we can close and ignore this PR.

p0pr0ck5:
Indeed this is not necessary

PR #106, opened 2017-09-06
Disable local cache

cilindrox:
`apk` provides a handy `--no-cache` flag that [disables the global cache][1]. This is favored over `rm -rf /var/cache/apk/*`

Replaces the `MAINTAINER` instruction, which has been [deprecated][2], with a `LABEL`.

[1]: https://github.com/gliderlabs/docker-alpine/tree/master#usage
[2]: https://docs.docker.com/engine/reference/builder/#maintainer-deprecated


salazar:
@cilindrox Thanks for your contribution!

PR #111, opened 2017-09-21
feat(enterprise): modify the dockerfile so we can build the image from an apk on the filesystem

hutchic:


hutchic:
[(3) Previewer should run Kong EE](https://trello.com/c/CwxQyetU/172-3-previewer-should-run-kong-ee)

kennethklee:
is the purpose for making the `Dockerfile` read a "kong.tgz" that would be generic enough for building both kong-ee and kong-ce?

because the makefile downloads only the ce version.

hutchic:
the Makefile only downloads the CE on purpose since we can't / won't expose details on how to build a kong-ee publicly (which this repository is) those details are in a private repository

PR #118, opened 2017-10-24
chore(*) split centos and alpine into separate directories

p0pr0ck5:


p0pr0ck5:
Based on feedback from https://github.com/docker-library/official-images/pull/3619

PR #120, opened 2017-10-27
LTS support for 0.10.4

p0pr0ck5:


lts/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM centos:7
+MAINTAINER Marco Palladino, marco@mashape.com
+
+ENV KONG_VERSION 0.10.4
+
+RUN yum install -y wget https://bintray.com/kong/kong-community-edition-rpm/download_file?file_path=dists%2Fkong-community-edition-$KONG_VERSION.el7.noarch.rpm && \
+    yum clean all
+
+RUN wget -O /usr/local/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v1.1.3/dumb-init_1.1.3_amd64 && \
```
salazar: We could avoid multiple RUN commands -- it's an optimization for size; instead, use && to put multiple commands together

p0pr0ck5: This is a copy paste from https://github.com/Kong/docker-kong/blob/a209825a9a74f9921c71c13ecb6e39a1b8e18aef/Dockerfile; I'd like to avoid any unnecessary changes to an older release if possible?

salazar: OK with me.

Issue #124, opened 2017-11-28
[alpine] Unable to set Timezone


bluemalkin:
In Docker it is standard practice to pass an environment variable TZ=timezone to set the desired timezone in the container.

This does not work with the alpine Docker image because it's missing the tzdata package.

Can we please have this added to the Dockerfile ?

Thanks !

bluemalkin:
PR raised - please approve ! 😃

salazar:
Closing this - #125 has been merged. Thanks @bluemalkin!

bluemalkin:
Thanks @salazar ! When can we expect this to be released please ?

salazar:
@bluemalkin A new Kong version was released today. The Docker image is already up - check Docker store!

PR #125, opened 2017-11-28
Added tzdata Alpine package

bluemalkin:
Adding support for TZ environment variable in alpine as per #124


salazar:
@bluemalkin Thanks for the contribution - merging it!

PR #134, opened 2018-01-11
Update Dockerfile

prolic:


hutchic:
thanks for the PR we've since updated to no longer use containerpilot and nginx in this specific way

PR #148, opened 2018-03-21
Docker STOPSIGNAL: Use SIGQUIT instead of SIGTERM

cyrilgdn:
What:
-------

Kong container should stop Nginx with a SIGQUIT signal instead of SIGTERM.

Why:
------

From Nginx man page:
```
SIGINT, SIGTERM  Shut down quickly.
SIGQUIT          Shut down gracefully.
```

Use case:

We have 2 instances of Kong running in our environments and we want to upgrade them (either for a Kong upgrade, or just some configurations changes).

Our orchestrator will do a rolling upgrade, for that it will:
 * Stop sending requests to the first instance
 * docker stop this instance
 * Starting a new one
 * Sending traffic to this new instance
Then it will do the same for the second instance.

So with this process, we have no downtime on our API. But at the docker stop step, with a SIGTERM, Nginx will just cancel all active connections so we have some errors on the associated requests.

With a SIGQUIT, Nginx will stop process new requests but will let ending the actives ones, so we'll have no errors on APi calls.

### DIFF
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -21,6 +21,6 @@ ENTRYPOINT ["/docker-entrypoint.sh"]
 
 EXPOSE 8000 8443 8001 8444
 
-STOPSIGNAL SIGTERM
+STOPSIGNAL SIGQUIT
 
 CMD ["/usr/local/openresty/nginx/sbin/nginx", "-c", "/usr/local/kong/nginx.conf", "-p", "/usr/local/kong/"]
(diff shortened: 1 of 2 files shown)


Tieske:
orchestration tools (k8s) send signals as well, they do not necessairly use the docker commands. By changing this, the behaviour would deviate in those circumstances. Container configured for graceful exit, but k8s sending sigterm anyway, bypassing that setting.

So imo it is better to be explicit in doing it the same always, and the user gets the same consistent (not necessarily most convenient) behavior.

Tieske:
see also: https://github.com/Kong/docs.konghq.com/pull/1040

PR #149, opened 2018-03-22
chore(*) bump to Kong CE 0.13.0

thibaultcha:


alpine/Dockerfile
```diff
@@ -1,8 +1,8 @@
 FROM alpine:3.6
 LABEL maintainer Marco Palladino, marco@mashape.com
```
hishamhm: are we keeping this as @mashape.com? (same below)

thibaultcha: Ideally not - also not an individual, this should be a team or support email. Pinged @salazar about it before, I think this got queued up

salazar: My personal opinion is that we should either use our team alias or, if it's not supposed to be made public, a team-specific support address.

Issue #151, opened 2018-03-27
Compose: Kong Dockerfile fails


iwittkau:
In `compose/kong/Dockerfile` the `curl` command fails due to the absence of a curl binary in the `kong` image: [Dockerfile:6](https://github.com/Kong/docker-kong/blob/914bba64bc7e1580b0e7e64e082459ca0c4e7eb7/compose/kong/Dockerfile#L6).

To try out the docker composition I added 

```Dockerfile
RUN apk add --no-cache curl
```

to the dockerfile on my machine, which resolves the issue temporally.

etsrepo:
https://github.com/Kong/docker-kong/pull/152

hutchic:
I believe this is resolved

PR #154, opened 2018-04-06
add curl into the alpine kong image

chenglx:
without curl, when we run docker-compose up, it will show errors like /bin/sh: curl: not found


chenglx:
for issue #151

PR #166, opened 2018-06-29
fix: do not unconditionally set KONG_PREFIX env

thibaultcha:
This PR fixes several flaws with the previous implementation:

* we do not unconditionally set KONG_PREFIX as an env variable
* we only set KONG_NGINX_DAEMON (to `off`) if we are starting Kong
  itself
* we use a safer way to determine whether or not we should prepare the
  prefix (only when `CMD` is `"kong (re)start"`)

It is now possible to use another prefix than `/usr/local/kong` within
the container (by specifying `-e KONG_PREFIX=`) which can be necessary
for specific container platforms (e.g. OpenShift).

### DIFF
--- a/alpine/docker-entrypoint.sh
+++ b/alpine/docker-entrypoint.sh
@@ -1,15 +1,14 @@
 #!/bin/sh
 set -e
 
-# Disabling nginx daemon mode
-export KONG_NGINX_DAEMON="off"
+if [[ "$1" == "kong" && "$2" =~ "start" ]]; then
+  export KONG_NGINX_DAEMON=off
+  PREFIX=${KONG_PREFIX:=/usr/local/kong}
 
-# Setting default prefix (override any existing variable)
-export KONG_PREFIX="/usr/local/kong"
-
-# Prepare Kong prefix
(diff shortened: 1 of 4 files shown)


alpine/docker-entrypoint.sh
```diff
@@ -1,15 +1,14 @@
...
-# Disabling nginx daemon mode
-export KONG_NGINX_DAEMON="off"
-
-# Setting default prefix (override any existing variable)
-export KONG_PREFIX="/usr/local/kong"
-
-# Prepare Kong prefix
-if [ "$1" = "/usr/local/openresty/nginx/sbin/nginx" ]; then
-	kong prepare -p "/usr/local/kong"
+if [[ "$1" == "kong" && "$2" =~ "start" ]]; then
```
Tieske: shouldn't this be `"$2" == "start"` ?

No shell guru, but doesn't this enable daemon mode on every-but-start command, instead of only start?

thibaultcha: The intent is to run this branch on `start` and `restart`. See #167 for the fix in which the intent is clearer, which allows us to stick to bourne shell.

PR #168, opened 2018-07-03
fix(entrypoint) allow the use of --run-migrations

thibaultcha:


yhourdel:
@thibaultcha could you please describe how you then use --run-migrations, please?

thibaultcha:
@yhourdel Ideally, you should not use it. Usage of this flag is disabled since a single node should be running migrations as per the current migrations mechanism. This flag is supported for _legacy_ reason, so if you are not using it, better avoid doing so.

yhourdel:
OK thanks, so we'll stick with the standard template with 2 nodes.

PR #172, opened 2018-07-22
Added support for OpenShift

bartoszj:
From OpenShift [docs](https://docs.openshift.org/3.9/creating_images/guidelines.html#use-uid):

> By default, OpenShift Origin runs containers using an arbitrarily assigned user ID. This provides additional security against processes escaping the container due to a container engine vulnerability and thereby achieving escalated permissions on the host node.

I've added small changes, so Kong could be run on OpenShift out of box.

Please also add those changes to Enterprise version.

### DIFF
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -14,7 +14,13 @@ RUN apk add --no-cache --virtual .build-deps wget tar ca-certificates \
 	&& rm -rf /tmp/usr \
 	&& cp -R /tmp/etc / \
 	&& rm -rf /tmp/etc \
-	&& apk del .build-deps
+	&& apk del .build-deps \
+	# OpenShift specific. OpenShift runs containers using an arbitrarily assigned user ID.
+	# This user doesn't have access to change file permissions during runtime, they have to be changed during image building.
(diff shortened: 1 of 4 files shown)


hutchic:
this seems mostly sensible.

Is there any reason to not `chmod` to a specific UID/GID inside the entrypoint script? 

ie if the ENV GID and UID exists `chmod -r` the files otherwise leave them.

bartoszj:
Unfortunately, OpenShift runs containers using an arbitrarily assigned user ID. This user doesn't have permissions to change file permissions during runtime, so they have to be changed during image building. The container user is always a member of the root group os it is safe to change GID to `0`.
At this [link](https://docs.okd.io/latest/creating_images/guidelines.html#use-uid) you can find more information.

Tieske:
hit this when browsing: maybe include a comment to explain why this is done, and that it is OpenShift specific, to save future us from a lot of head-scratching

bartoszj:
Hello. I've added comments to Dockerfiles

Tieske:
@hutchic lgtm

hutchic:
naive question: why is this change only relevant to the alpine and centos images?

bartoszj:
Sorry. I've somehow missed RedHat and legacy images :(

PR #186, opened 2018-10-27
fix(README) Correct link to docs site

jeveleth:


jeveleth:
Good call, @gszr. Updated.

thibaultcha:
Ah, I see this PR from the commit; it seems like the commits were not squashed at merge time. We try to have a linear git history in which similar changes are grouped together within the same commit :)

PR #187, opened 2018-10-29
fix(compose) use kong user for pg healthcheck

rainest:
Use the `kong` user for Postgres healthchecks instead of `postgres`. An [upstream fix to the Postgres image](https://github.com/docker-library/postgres/commit/3f585c58df93e93b730c09a13e8904b96fa20c58) removes the default `postgres` user if another user is specified.

Fix #180
Fix #179
Fix #174

PR #189, opened 2018-10-30
set the default PG_USER in .env

hutchic:
Inspired by https://github.com/Kong/docker-kong/pull/187/files set a sensible default for the postgres user which can be overridden by exporting `PG_USER` as an environment variable

### DIFF
--- a/compose/.env
+++ b/compose/.env
@@ -1,0 +1,1 @@
+PG_USER=kong
--- a/compose/.env
+++ b/compose/.env
@@ -1,1 +1,0 @@
-PG_USER=kong
--- a/compose/docker-compose.yml
+++ b/compose/docker-compose.yml
@@ -3,10 +3,10 @@ services:
   kong-database:
     image: postgres:9.5
     environment:
-      - POSTGRES_USER=${PG_USER}
+      - POSTGRES_USER=${PG_USER:-kong}
       - POSTGRES_DB=kong
     healthcheck:
-      test: ["CMD", "pg_isready", "-U", "${PG_USER}"]
+      test: ["CMD", "pg_isready", "-U", "${PG_USER:-kong}"]
       interval: 30s
       timeout: 30s
       retries: 3
@@ -21,6 +21,7 @@ services:
       - KONG_DATABASE=postgres
       - KONG_PG_HOST=kong-database
       - KONG_PG_DATABASE=kong
+      - KONG_PG_USER=${PG_USER:-kong}
       - KONG_ADMIN_LISTEN=0.0.0.0:8001        
     ports:
       - "8000:8000"
--- a/compose/docker-compose.yml
+++ b/compose/docker-compose.yml
@@ -3,10 +3,10 @@ services:
   kong-database:
     image: postgres:9.5
     environment:
-      - POSTGRES_USER=kong
+      - POSTGRES_USER=${PG_USER}
       - POSTGRES_DB=kong
     healthcheck:
-      test: ["CMD", "pg_isready", "-U", "postgres"]
+      test: ["CMD", "pg_isready", "-U", "${PG_USER}"]
       interval: 30s
       timeout: 30s
       retries: 3


chris-scentregroup:
This is my result using this compose now:
```
$ docker-compose up
WARNING: The PG_USER variable is not set. Defaulting to a blank string.
Recreating compose_kong-database_1 ... done
Recreating compose_kong_1          ... done
Attaching to compose_kong-database_1, compose_kong_1
kong-database_1  | LOG:  database system was interrupted; last known up at 2018-11-01 05:40:43 UTC
kong-database_1  | LOG:  database system was not properly shut down; automatic recovery in progress
kong-database_1  | LOG:  invalid record length at 0/183A728
kong-database_1  | LOG:  redo is not required
kong-database_1  | LOG:  MultiXact member wraparound protections are now enabled
kong-database_1  | LOG:  autovacuum launcher started
kong-database_1  | LOG:  database system is ready to accept connections
kong-database_1  | FATAL:  role "root" does not exist
kong-database_1  | FATAL:  role "root" does not exist
kong-database_1  | FATAL:  role "root" does not exist
```

chris-scentregroup:
Perhaps we can solve not needing the `.env` by doing defaults like this:
```
services:
  kong-database:
    image: postgres:9.5
    environment:
      - POSTGRES_USER=${PG_USER:-kong}
      - POSTGRES_DB=kong
    healthcheck:
      test: ["CMD", "pg_isready", "-U", "${PG_USER:-kong}"]
      interval: 30s
      timeout: 30s
      retries: 3
    restart: on-failure
```

hutchic:
@chris-scentregroup that's odd 🤔 

The docker-compose manual is fairly clear that `.env` should set the environment variable in the absence of you having one. https://docs.docker.com/compose/environment-variables/#the-env-file

It works for me on the following setup:
```
 docker version
Client:
 Version:      17.05.0-ce
 API version:  1.29
 Go version:   go1.7.5
 Git commit:   89658be
 Built:        Thu May  4 22:10:54 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.05.0-ce
 API version:  1.29 (minimum version 1.12)
 Go version:   go1.7.5
 Git commit:   89658be
 Built:        Thu May  4 22:10:54 2017
 OS/Arch:      linux/amd64
 Experimental: false
ubuntu@ip-172-31-24-121:~/projects/mashape/docker-kong/compose$ docker-compose version
docker-compose version 1.21.2, build a133471
docker-py version: 3.3.0
CPython version: 3.6.5
OpenSSL version: OpenSSL 1.0.1t  3 May 2016
ubuntu@ip-172-31-24-121:~/projects/mashape/docker-kong/compose$ uname -a
Linux ip-172-31-24-121 4.4.0-1069-aws #79-Ubuntu SMP Mon Sep 24 15:01:41 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
ubuntu@ip-172-31-24-121:~/projects/mashape/docker-kong/compose$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.3 LTS
Release:        16.04
Codename:       xenial
```

try nuking the existing containers if you can and giving it a go from a clean slate `docker-compose rm -f`? Otherwise if your suggestion works as well we'll go with it ¯\\_(ツ)_/¯

chris-scentregroup:
@hutchic shouldn't we be making sure it works out of the box? My solution saves the need to create a `.env` and a dotfile should be hidden and not included in a source repo.

hutchic:
@chris-scentregroup absolutely which is why the `.env` file is committed to the repository so it (at least for me locally) does work out of the box

ref: https://github.com/Kong/docker-kong/blob/fix/configurable-db-user/compose/.env

chris-scentregroup:
`.env` is typically git ignored as its meant to be a local file where users can customise values. If its stored in the repo, user customises values in the file, adds other variables etc. and it comes back as modified.
We should be using defaults per my example or if we do include the `.env` simply use `POSTGRES_USER` saving the need for a worker variable.

chris-scentregroup:
Having success with this https://github.com/flaccid/compose-kong/blob/master/ce/docker-compose.yml where there is no need for a worker variable and does not result in `FATAL:  role "root" does not exist` errors.

hutchic:
@chris-scentregroup doesn't look that different from this PR but still an improvement in spots. 

PR it and I'll merge that and close this one 👍

chris-scentregroup:
Its quite different really, small changes can equal big changes.

Raised https://github.com/Kong/docker-kong/pull/193.

hutchic:
used https://github.com/Kong/docker-kong/pull/193 instead

Issue #181, opened 2018-10-05
Error when running with custom ngnix template


Mourdraug:
I can't get container to run with even simpliest template.

### docker-compose.yml
```
version: '3'

volumes:
  kong_db_test:
    driver: local

services:
  kong-postgres:
    image: postgres:9.5-alpine
    volumes:
      - kong_db_test:/var/lib/postgresql/data
    environment:
      POSTGRES_DB: kong
      POSTGRES_USER: kong
      POSTGRES_PASSWORD: e?~&yT0?ts397$wb.4;+
    healthcheck:
      test: ['CMD', 'pg_isready', '-U', 'kong']
      interval: 5s
      timeout: 5s
      retries: 5
  kong-migration:
    image: kong:0.14.1-alpine
    command: 'kong migrations up'
    restart: on-failure
    environment:
      KONG_DATABASE: postgres
      KONG_PG_HOST: kong-postgres
      KONG_PG_PORT: 5432
      KONG_PG_DATABASE: kong
      KONG_PG_USER: kong
      KONG_PG_PASSWORD: e?~&yT0?ts397$wb.4;+
      KONG_LOG_LEVEL: info
    depends_on:
      - kong-postgres
  kong:
    build: ./kong
    environment:
      KONG_DATABASE: postgres
      KONG_PG_HOST: kong-postgres
      KONG_PG_PORT: 5432
      KONG_PG_DATABASE: kong
      KONG_PG_USER: kong
      KONG_PG_PASSWORD: e?~&yT0?ts397$wb.4;+
      KONG_LOG_LEVEL: info
    ports:
      - 8000:8000
    depends_on:
      - kong-postgres
      - kong-migration

```
### ./kong/Dockerfile

    FROM kong:0.14.1-alpine

    COPY . /etc/kong/

    CMD kong start --nginx-conf /etc/kong/nginx.template
### ./kong/ngnix.template
```
worker_processes ${{NGINX_WORKER_PROCESSES}}; # can be set by kong.conf
daemon ${{NGINX_DAEMON}};                     # can be set by kong.conf

pid pids/nginx.pid;                      # this setting is mandatory
error_log logs/error.log ${{LOG_LEVEL}}; # can be set by kong.conf

events {
    use epoll; # custom setting
    multi_accept on;
}

http {
    # include default Kong Nginx config
    include 'nginx-kong.conf';
}
```
### Log
```
[...]

PR #191, opened 2018-11-01
Pass arguments to 'kong prepare' in docker-entrypoint.sh

fooka03:
The idea here is to allow additional arguments to pass through to the `kong prepare` command in the entrypoint script without fundamentally changing how the container operates normally.  This way, one could supply flags for verbose/debug output, or a custom nginx template (As is the case for Issue #181 )

I've tested this with the alpine container, but I imagine there wouldn't be any issues with this change for centos/rhel either.


Tieske:
@fooka03 this is a nice change. Would you mind fixing the merge conflict?

fooka03:
@Tieske conflicts fixed as requested.

Tieske:
@hutchic can we merge this?

PR #192, opened 2018-11-06
Bumbed compose version, fixed user, and passed all params

vegasbrianc:
Updated the compose:
- Bumped compose version to support new functionality
- added network kone-net
- added volume kong_data for DB state

### DIFF
--- a/compose/docker-compose.yml
+++ b/compose/docker-compose.yml
@@ -1,4 +1,10 @@
 version: '2.1'
+
+volumes:
+    kong_data: {}
+
+networks:
+  kong-net:
 services:
   kong:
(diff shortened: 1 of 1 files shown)


hutchic:
we've heavily updated `docker-compose.yml` recently. The changes include some of the improvements you have proposed.

Update this PR so it's only the network and volume would still be good improvements 👍 

If we do upgrade `docker-compose` to `version: '3.1'` we'll need to double check if it's necessary to add a `restart: always` to Kong with the (unfortunate) deprecation of `depends_on`

hutchic:
Thanks @vegasbrianc 👍

PR #193, opened 2018-11-08
Various compose improvements

flaccid:
- alternative/better solution for pull #189 
- add use of postgres password for db service
- add in some more env vars for kong

### DIFF
--- a/compose/docker-compose.yml
+++ b/compose/docker-compose.yml
@@ -1,6 +1,6 @@
 version: '2.1'
 services:
-  kong-ce:
+  kong:
     image: "${KONG_DOCKER_TAG:-kong:latest}"
     command: kong start --run-migrations
     depends_on:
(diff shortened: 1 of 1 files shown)


compose/docker-compose.yml
```diff
@@ -1,31 +1,42 @@
 version: '2.1'
 services:
-  kong-database:
+  kong-ce:
```
hutchic: can we leave this as `kong:`?

hutchic:
If we are we should add that to the repository somewhere or cross-link to https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#commit-message-format

thibaultcha:
We most definitely are, as pointed out already in a few commits that made it to master last week. 

The lack of contributing file is not really an excuse, as the format is project-wide policy, and a quick glance at the commit history is a strong hint that the format is used in the repository as well.

PR #194, opened 2018-11-12
feat(exit) add script for graceful exit

Tieske:
This adds a script to do a graceful exit. It takes 2 parameters, pre-timeout and post-timeout.

TLDR:

 - pre-timeout allows to drain NEW connections coming in
 - post-timeout allows to drain ACTIVE connections to completion

The pre-timeout (default 3 seconds) is the wait period before the actual exit is executed. This is useful for orchestration tools (👀 at you k8s!), that are eventually consistent in their updates. Without the wait period, Kong will stop accepting new connections before the loadbalancers and other components are updated and actually stop sending new connections, which causes lost connections.

The post-timeout is the standard timeout used for the `kong quit` command.

So using this script in Kubernetes allows to set a PreStop hook:
```yaml
            preStop:
              exec:
                command: ["/graceful-exit.sh", [3], [10]]
```

NOTE: untested, for discussion only (also needs updates to the others if accepted)

### DIFF
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -17,6 +17,7 @@ RUN apk add --no-cache --virtual .build-deps wget tar ca-certificates \
 	&& apk del .build-deps
 
 COPY docker-entrypoint.sh /docker-entrypoint.sh
+COPY graceful-exit.sh /graceful-exit.sh
 ENTRYPOINT ["/docker-entrypoint.sh"]
 
 EXPOSE 8000 8443 8001 8444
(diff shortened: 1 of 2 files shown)


james-callahan:
What made you pick 3 seconds?
I'm always wary of adding sleeps into init scripts: they're a source of all sorts of "that was weird" bugs where "restart it" fixes it.
Is there some event we could wait on instead?

Tieske:
no event that I know of. It would have to be something like "x amount of time that Kong did not receive any new connections", but that sounds awful

Tieske:
@hutchic can you have a look? if this is something we want, we should update it, if not, feel free to close. Essentially it was just for discussion purposes anyway. So let's decide and move on.

Tieske:
since https://github.com/Kong/kong/pull/4201 introduced a `--wait` flag, this is no longer needed.

PR #199, opened 2018-12-13
feat(mesh) run Kong as non-root user

hutchic:


alpine/Dockerfile
```diff
@@ -4,7 +4,10 @@ LABEL maintainer="Kong Core Team <team-core@konghq.com>"
 ENV KONG_VERSION 1.0.0rc4
 ENV KONG_SHA256 b65c7c01a4d523b1687898c05793aae71645e8ba0b89d59343c157f4f88db7dd
 
-RUN apk add --no-cache --virtual .build-deps wget tar ca-certificates \
+RUN adduser -Su 1337 kong \
+  && mkdir -p "/usr/local/kong" \
```
james-callahan: indentation here is spaces vs tabs for following lines.

alpine/Dockerfile
```diff
@@ -4,7 +4,10 @@ LABEL maintainer="Kong Core Team <team-core@konghq.com>"
 ENV KONG_VERSION 1.0.0rc4
 ENV KONG_SHA256 b65c7c01a4d523b1687898c05793aae71645e8ba0b89d59343c157f4f88db7dd
 
-RUN apk add --no-cache --virtual .build-deps wget tar ca-certificates \
+RUN adduser -Su 1337 kong \
+	&& mkdir -p "/usr/local/kong" \
```
james-callahan: As in the other dockerfiles this `mkdir` can be removed and the `chown` moved below the extraction.

hutchic: nope we can't

```
kong.tar.gz: OK
chown: /usr/local/kong: No such file or directory
```

james-callahan: huh. okay.....

PR #205, opened 2018-12-21
feat(plugins) enables adding custom plugins/template

Tieske:
new dockerfile that extends an existing Kong image by injecting custom plugins (with all their dependencies) and/or a custom template in a new image.

Not sure this is the right place, or if I should create a separate repository for it...


Tieske:
@hutchic By now I actually think this PR belongs in this repo. This is the goto reference for creating docker files, and as such the customization  belong here imo.

Can we merge this?

Tieske:
@hutchic ping

hutchic:
I think going a different direction then this would be preferable (specifically git://kong-build-tools `make development`) but this isn't causing any regression or breaking change so let's go ahead and merge it regardless 👍

Tieske:
@hutchic just to be on the same page here; this is not for development. This is for creating production images, with best-practices Kong images under the hood, just extended with customizations. Packaged as a dockerfile so it makes it easy to include it in a CI/CD pipeline.

So imo a different purpose than the Kong build tools

hutchic:
@Tieske yup ok -- reread the README.md and now I get it let's do it 👍

PR #206, opened 2018-12-21
fix(setcap)

hutchic:
moves setcap out of Dockerfile and into the docker entrypoint's

### DIFF
--- a/alpine/docker-entrypoint.sh
+++ b/alpine/docker-entrypoint.sh
@@ -7,9 +7,15 @@ if [[ "$1" == "kong" ]]; then
   PREFIX=${KONG_PREFIX:=/usr/local/kong}
 
   if [[ "$2" == "docker-start" ]]; then
-    kong prepare -p "$PREFIX"
+    chown kong "$PREFIX"
+    su-exec kong kong prepare -p "$PREFIX"
+    
+    setcap cap_net_raw=+ep /usr/local/openresty/nginx/sbin/nginx
 
-    exec /usr/local/openresty/nginx/sbin/nginx \
+    chmod 777 /proc/self/fd/1
+    chmod 777 /proc/self/fd/2
+
+    su-exec kong /usr/local/openresty/nginx/sbin/nginx \
       -p "$PREFIX" \
       -c nginx.conf
   fi
--- a/alpine/docker-entrypoint.sh
+++ b/alpine/docker-entrypoint.sh
@@ -3,19 +3,30 @@ set -e
 
 export KONG_NGINX_DAEMON=off
 
+has_transparent() {
+  echo "$1" | grep -E "[^\s,]+\s+transparent\b" >/dev/null
+}
+
 if [[ "$1" == "kong" ]]; then
   PREFIX=${KONG_PREFIX:=/usr/local/kong}
 
   if [[ "$2" == "docker-start" ]]; then
-    chown kong "$PREFIX"
-    su-exec kong kong prepare -p "$PREFIX"
-    
-    setcap cap_net_raw=+ep /usr/local/openresty/nginx/sbin/nginx
+    kong prepare -p "$PREFIX"
+    chown -R kong "$PREFIX"
 
-    chmod 777 /proc/self/fd/1
-    chmod 777 /proc/self/fd/2
+    # workaround for https://github.com/moby/moby/issues/31243
+    chmod o+w /proc/self/fd/1
+    chmod o+w /proc/self/fd/2
 
-    su-exec kong /usr/local/openresty/nginx/sbin/nginx \
+    if [ ! -z ${SET_CAP_NET_RAW} ] \
+        || has_transparent "$KONG_STREAM_LISTEN" \
+        || has_transparent "$KONG_PROXY_LISTEN" \
+        || has_transparent "$KONG_ADMIN_LISTEN";
+    then
+      setcap cap_net_raw=+ep /usr/local/openresty/nginx/sbin/nginx
+    fi
+
+    exec su-exec kong /usr/local/openresty/nginx/sbin/nginx \
       -p "$PREFIX" \
       -c nginx.conf
   fi
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -1,8 +1,8 @@
 FROM alpine:3.6
 LABEL maintainer="Kong Core Team <team-core@konghq.com>"
 
-ENV KONG_VERSION 1.0.0
-ENV KONG_SHA256 79001d96c28d774bcf868f281e8e3486c4027ed35d69da6560172a742240b862
+ENV KONG_VERSION 0.15.0
+ENV KONG_SHA256 f2adc4e40a4a33c657955d6d9ec034d80827f915bb9fe8c2ac2a36aed7edf13b
 
 RUN adduser -Su 1337 kong \
 	&& mkdir -p "/usr/local/kong" \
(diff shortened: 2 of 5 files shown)


alpine/docker-entrypoint.sh
```diff
@@ -7,9 +7,15 @@ if [[ "$1" == "kong" ]]; then
   PREFIX=${KONG_PREFIX:=/usr/local/kong}
 
   if [[ "$2" == "docker-start" ]]; then
-    kong prepare -p "$PREFIX"
+    chown kong "$PREFIX"
+    su-exec kong kong prepare -p "$PREFIX"
+    
+    setcap cap_net_raw=+ep /usr/local/openresty/nginx/sbin/nginx
```
hutchic: need to wrap this in an `IF`

alpine/docker-entrypoint.sh
```diff
@@ -7,9 +7,17 @@ if [[ "$1" == "kong" ]]; then
   PREFIX=${KONG_PREFIX:=/usr/local/kong}
 
   if [[ "$2" == "docker-start" ]]; then
-    kong prepare -p "$PREFIX"
+    chown kong "$PREFIX"
+    su-exec kong kong prepare -p "$PREFIX"
+    
+    if [ ! -z ${KONG_STREAM_LISTEN+x} ]; then
```
james-callahan: This check isn't correct. The transparent option gets used with http too.

alpine/docker-entrypoint.sh
```diff
@@ -7,9 +7,17 @@ if [[ "$1" == "kong" ]]; then
...
+      setcap cap_net_raw=+ep /usr/local/openresty/nginx/sbin/nginx
+    fi
 
-    exec /usr/local/openresty/nginx/sbin/nginx \
+    chmod 777 /proc/self/fd/1
```
james-callahan: This `chmod` is really weird. I think I remember watching you run into this issue, but it really shouldn't happen in the first place as far as I know...

james-callahan:
:( the `capsh` that ships with alpine 3.8 is too old to have the `--addamb` argument. Does anyone know of a program easily available that execs with new ambient capabilities?

alpine/docker-entrypoint.sh
hutchic: afraid I don't understand what you mean. TLDR what **should** the check and the command be keeping in mind alpine and centos compatibility?

james-callahan: There should be no `if` here.

hutchic: There needs to be an if of some sort. We only need to setcap for transparent

james-callahan: One condition could be if stream_listen, admin_listen or proxy_listen contain `transparent`. But we really just want it unconditional...

hutchic: I feel like we don't want it unless I'm unaware of some future plan we have

If I personally was running Kong I would want it's security profile restricted explicitly to what it requires and no more. So if I'm running with transparent then setcap is fine otherwise IMO it's an unnecessary privilege

Also the docker pr that lead to this suggested this route ...

subnetmarco: > One condition could be if stream_listen, admin_listen or proxy_listen contain transparent.

Can't add those conditions as well to make it more correct? @james-callahan

james-callahan: We can, though the simplest implementation of that check (checking if they contain the substring `transparent`) may have false positives if it is found in e.g. the host.
I feel like it isn't a huge issue to have the capability and not use it.

subnetmarco: Maybe a regular expression to check if it’s specified after the host will do it?

alpine/docker-entrypoint.sh
```diff
@@ -7,9 +7,17 @@ if [[ "$1" == "kong" ]]; then
...
+    chmod 777 /proc/self/fd/2
+
+    su-exec kong /usr/local/openresty/nginx/sbin/nginx \
```
james-callahan: This should be `exec su-exec`. otherwise the script will continue.

alpine/docker-entrypoint.sh
james-callahan: Yes that is possible; though you will need to repeat the check for each of the listen variables, and to make sure it handles the comma separated array form correctly.

subnetmarco: For your convenience this is regex that worked for me: `((?!transparent))\s+transparent`

It uses a lookahead to prevent matching any hostname that includes the word `transparent` in it, and only matches the actual `transparent` property after the hostname.

subnetmarco: You can play with the Regex here: https://www.regexpal.com/?fam=106839

james-callahan: That fails on e.g. `localhost:3333, transparent:4444, bar:5555 ssl`

subnetmarco: How about this: `^([A-Za-z0-9:,\s\.]*\s+)transparent(\s+[\sA-Za-z0-9_]+)*$`

If it fully matches, then `transparent` is set, otherwise it's not: https://www.regexpal.com/?fam=106862

subnetmarco: This regex seems to be working for me: `^([A-Za-z0-9:,\s\.]+\s+)transparent((\s+[A-Za-z0-9_]*)|(\s*,[A-Za-z0-9:,\s\.]+))*$` ([test it here]( https://www.regexpal.com/?fam=106863)).

thibaultcha: This simplified version works and detects tricky use-cases as well: `[^\s,]+\s+transparent\b`

E.g.
* `host:2313 ssl transparent_not, host:8000 transparent_not, host:9000 transparent, host:7000 transparent`
* `   transparent, host:7000 transparent`

It is also able to determine how many instances of `transparent` there are, and extract their corresponding listener (although we don't care in this case).

alpine/docker-entrypoint.sh
thibaultcha: That shouldn't be necessary (https://github.com/ncopa/su-exec/blob/master/su-exec.c#L103). It is also documented in the project's README: https://github.com/ncopa/su-exec#tty--parentchild-handling

james-callahan: Unless you `exec` the bash process will live on as a parent.
The readme there is saying su-exec won't stick around; not the parent of su-exec.

thibaultcha: Indeed. Also, any thoughts on https://github.com/ncopa/su-exec/issues/23?

thibaultcha:
New version pushed. Please see the commit message for details. This has been delayed long enough, and the current implementation already is an improvement over the existing image, plus supports auto-detection of the transparent option as the best compromise we could find.

hutchic:
I tested the latest push in docker-compose and minikube both environments and setups worked as expected 👍

Tieske:
shouldn't the same changes be applied to the rhel image?

james-callahan:
> shouldn't the same changes be applied to the rhel image?

We don't have any way to test rhel.

subnetmarco:
Wouldn't this apply to any other distribution too (Ubuntu, Debian, CentOS in addition to RHEL, etc)?

james-callahan:
> Wouldn't this apply to any other distribution too (Ubuntu, Debian, CentOS in addition to RHEL, etc)?

I'm not sure what you mean by this?

The method is different per distro (su-exec is not that common). This repository only creates docker images for alpine, centos and rhel. rhel is the only one of those for which we have no testing setup.

subnetmarco:
What are the next steps to support this feature on any other Docker image we create? @Tieske is right in his observation.

hutchic:
@Kong/team-fast-track handles the RHEL container but I believe the RHEL container is loosely based on our centos image so it might be fine but probably best to follow up with folks on that team

PR #209, opened 2019-01-01
Fix kong migrations command

abdul:
The `kong migrations bootstrap` doesn't work anymore. Update the command to `kong migrations up`.


Tieske:
duplicate of #201

PR #210, opened 2019-01-08
Kong 1.0.0 image based on debian:stretch

fazlerabby7:


hutchic:
without a strong reason to do this I'm leaning towards not merging it. 

We have an alpine and a centos that we officially support so I'm not convinced we also need a debian. Ist there a specific reason you need / want this. If it's just for familiarity of debian (which I can appreciate) I'd say make a github repository of your own with this Dockerfile and setup docker hub to auto build the image for you.

PR #213, opened 2019-01-16
setcap CAP_NET_BIND_SERVICE to support deploying Kong to port 80 in h…

JoyZhang66:
I tried to deploy Kong1.0 to port 80 in docker with host network like below, but getting nginx: [emerg] bind() to 0.0.0.0:80 failed (13: Permission denied) error message.

docker run -d --network host --name kong -e "KONG_DATABASE=cassandra" -e "KONG_CASSANDRA_CONTACT_POINTS=XX.XX.XX.XX" -e "KONG_ADMIN_LISTEN=0.0.0.0:8001, 0.0.0.0:8444 ssl" -e "KONG_PROXY_LISTEN=0.0.0.0:80, 0.0.0.0:443 ssl" kong:1.0

I checked the code and found only cap_net_raw was set but CAP_NET_BIND_SERVICE not, so I added setcap CAP_NET_BIND_SERVICE to support deploying Kong to port 80 in host network.

CAP_NET_BIND_SERVICE
Bind a socket to Internet domain privileged ports (port numbers less than 1024).

CAP_NET_RAW
* Use RAW and PACKET sockets;
* bind to any address for transparent proxying.


alpine/docker-entrypoint.sh
```diff
@@ -26,6 +26,8 @@ if [[ "$1" == "kong" ]]; then
       setcap cap_net_raw=+ep /usr/local/openresty/nginx/sbin/nginx
     fi
 
+    setcap CAP_NET_BIND_SERVICE=+ep /usr/local/openresty/nginx/sbin/nginx
```
james-callahan: use lower case for consistency with the previous setcap command.

james-callahan: This breaks the above `setcap` command: it does not *add* a capability, but *set* the capabilities of the program.
You'll need to collect all the capabilities wanted together and do it in one setcap call.

alpine/docker-entrypoint.sh
```diff
@@ -23,7 +23,7 @@ if [[ "$1" == "kong" ]]; then
         || has_transparent "$KONG_PROXY_LISTEN" \
         || has_transparent "$KONG_ADMIN_LISTEN";
     then
-      setcap cap_net_raw=+ep /usr/local/openresty/nginx/sbin/nginx
+      setcap cap_net_raw,cap_net_bind_service=+ep /usr/local/openresty/nginx/sbin/nginx
```
james-callahan: You don't want this to be conditional on transparent.

JoyZhang66:
@james-callahan 
I didn't understand what you mean before, thank you for your update.

hutchic:
The test failure looks legit

```
BASE="centos" KONG_DOCKER_TAG="kong-centos" make build
cd compose
KONG_DOCKER_TAG="kong-centos" docker-compose up -d
docker-compose logs -f kong
Attaching to compose_kong_1
```

trollr:
When will this be merged?

hutchic:
Kong 1.5.0 and 2.0.1 no longer require setcap. It's good practice to avoid `setcap` unless absolutely required in official Docker images which prior to 1.4 was required to run Kong service mesh which has now been migrated to Kuma.

I would advise the following on a more recent Kong version or if your use case absolutely requires binding to 80 within the container making a custom docker image that is `FROM kong`
```
docker run -d --name kong \
     --network=kong-net \
     -e "KONG_DATABASE=cassandra" \
     -e "KONG_PG_HOST=kong-database" \
     -e "KONG_CASSANDRA_CONTACT_POINTS=kong-database" \
     -e "KONG_ADMIN_LISTEN=0.0.0.0:8001, 0.0.0.0:8444 ssl" \
     -p 8000:8000 \
     -p 8443:8443 \
     -p 80:8001 \
     -p 443:8443 \
     kong:2.0.1
```

diefans:
@hutchic this still adds `8443` to the `X-Forwarded-Port` and `Host` headers

hutchic:
@diefans I think your issue isn't related and might even belong on kong/kong as a github issue

diefans:
@hutchic I think indeed it is related but there are efforts in kong/kong already solving this problem: https://github.com/Kong/kong/issues/5018

Issue #224, opened 2019-02-01
Kong and Docker Swarm


OwenJacob:
### Summary

Official Docker Kong image does not make it through the entrypoint script when running as a service on Docker Swarm. docker logs for the container states:

```
chmod: /proc/self/fd/1: Permission denied
```
This is line 19 of the entrypoint for the image.

### Steps To Reproduce

1. Init a docker swarm
2. Use the docker-compose below
3. docker stack deploy kong
4. docker logs [container id]

### Additional Details & Logs

I have tried to build my own Docker file for Kong to workaround this issue. I used the official Debian Stretch container and installed the Kong deb package. This container starts Kong but then immediately restarts itself - no indication of why in the logs (I used kong start --vv in the entrypoint and I can see that Kong successfully starts). I wonder if anyone else has managed to get Kong working in Docker Swarm.

The networks are already created and working. The migration works and the tables exist in the DB.

docker-compose.yml:
```
version: '3'

services:
  db:
    image: postgres:9.6
    volumes:
      - ./postgres:/var/lib/postgresql/data
    env_file:
      - .env
    networks:
      - database-net

  migration:
    image: kong:latest
    command: kong migrations bootstrap
    env_file:
      - .env
    deploy:
      restart_policy:
        condition: on-failure
    networks:
      - database-net

  kong:
    image: kong:latest
    env_file:
      - .env
    deploy:
      restart_policy:
        condition: on-failure
        delay: 10s
        max_attempts: 3
      labels:
        - "traefik.enable=true"
        - "traefik.kongHTTP.port=8000"
        - "traefik.kongHTTPadmin.port=8001"
        - "traefik.kongHTTPS.port=8443"
        - "traefik.kongHTTPSadmin.port=8444"
        - "traefik.docker.network=swarm-net"
        - "traefik.frontend.rule=Host:api.gateway.swarm.sabay.com"
        - "traefik.kongHTTP.frontend.entryPoints=http8000"
        - "traefik.kongHTTPadmin.frontend.entryPoints=http8001"
        - "traefik.kongHTTPS.frontend.entryPoints=https8443"
        - "traefik.kongHTTPSadmin.frontend.entryPoints=https8444"
    networks:
      - swarm-net
      - database-net

networks:
  swarm-net:
    external: true
  database-net:
    external: true
```

.env
```
POSTGRES_USER=kong
POSTGRES_DB=kong
KONG_DATABASE=postgres
KONG_PG_HOST=db
[...]

PR #228, opened 2019-02-15
fix(docker-swarm) let the chmod command fail so docker-swarm is happy

hutchic:
A possible resolution to https://github.com/Kong/docker-kong/issues/224


hutchic:
No longer needed

PR #233, opened 2019-03-12
Changes to Alpine image to be run as anyuid (OpenShift)

svenwal:
This pull request changes the Dockerfile and entrypoint.sh in a way that the image can be run without root privileges and as an arbitrary user (anyuid). When those changes have been applied Kong can be run on a default OpenShift installation without changing any default settings regarding root user or arbitrary user.

See https://docs.openshift.com/enterprise/3.0/creating_images/guidelines.html for reference about the needed changes.

The Dockerfile changes in detail:
 - kong user is now member of the root group
 - the chown command which has been located in the entrypoint.sh until now is part of the Dockerfile (as we only have the permissions here to execute the command)
 - chgrp/chmod commands added so /usr/local/kong belongs to user kong's root group
 - USER kong added so image is run as kong from the start

The entrypoint.sh changes in detail:
 - chown and chgrp commands removed / relocated to Dockerfile (we are user kong in entrypoint.sh now so no permissions to do so anymore)
 - "exec su-exec kong" removed as we are king anyway

### DIFF
--- a/alpine/docker-entrypoint.sh
+++ b/alpine/docker-entrypoint.sh
@@ -13,11 +13,6 @@ if [[ "$1" == "kong" ]]; then
   if [[ "$2" == "docker-start" ]]; then
     shift 2
     kong prepare -p "$PREFIX" "$@"
-    chown -R kong "$PREFIX"
-
-    # workaround for https://github.com/moby/moby/issues/31243
-    chmod o+w /proc/self/fd/1
-    chmod o+w /proc/self/fd/2
 
     if [ ! -z ${SET_CAP_NET_RAW} ] \
         || has_transparent "$KONG_STREAM_LISTEN" \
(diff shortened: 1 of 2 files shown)


alpine/docker-entrypoint.sh
```diff
@@ -27,7 +22,7 @@ if [[ "$1" == "kong" ]]; then
       setcap cap_net_raw=+ep /usr/local/openresty/nginx/sbin/nginx
     fi
 
-    exec su-exec kong /usr/local/openresty/nginx/sbin/nginx \
+    /usr/local/openresty/nginx/sbin/nginx \
```
james-callahan: still need the `exec` or you get a bash process hanging around.

svenwal: I can tell you this container has been working fine for me (also for my Enterprise deployment) using docker-compose, Kubernetes and OpenShift. I do not have access to a full test environment so but use all the functions normally

hutchic:
I was super hopeful this would work and we could forgo the ugly workaround but appears it does not
```
standard_init_linux.go:207: exec user process caused "permission denied"
```

Specifically when running Kong as a transparent proxy to get around https://github.com/moby/moby/issues/31243 . If this works for your use case I'd encourage you to continue using it as is but the official Kong images need to maintain the capability to run as a transparent proxy.

It's on my todo list to make a test for this and also make the official container openshift compatible once again hopefully those two goals aren't in opposition :)

PR #234, opened 2019-03-12
Allow running as non-root

cassianoleal:
`su-exec` needs to run as root so as long as the `entrypoint` script uses it users cannot run the container as a non-root user.

This patch allows the user to set the `KONG_DO_NOT_SU_EXEC` flag to avoid running `su-exec` and allowing kong to be executed.

I'm running this on Kubernetes, mounting the `KONG_PREFIX` directory as an `emptyDir` and configuring Kong to run as `kong` user and mount `KONG_PREFIX` with `kong`'s GID:

```
      securityContext:
        fsGroup: 65533
        runAsUser: 1337
```

### DIFF
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -6,6 +6,7 @@ ENV KONG_SHA256 72abd186181b5ebb263c4e12db5d89cac529e2b5ca858015d44a34d560755b35
 
 RUN adduser -Su 1337 kong \
 	&& mkdir -p "/usr/local/kong" \
+	&& chown -R kong "/usr/local/kong" \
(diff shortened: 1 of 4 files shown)


cassianoleal:
There's still work to be done here. I have it running as non-root on my Kubernetes cluster with a slightly different config.

hutchic:
@cassianoleal thanks for taking a stab at this. Keep an eye out for gotchas around running Kong with transparent proxying turned on as others (including me) have ventured down this path before :) 

https://github.com/Kong/docker-kong/pull/233#issuecomment-472576912

hutchic:
This should be resolved now with https://github.com/Kong/docker-kong/pull/243 being merged (haven't updated the official images yet)

PR #236, opened 2019-03-15
OpenShift deployment with anyuid / arbitrary user and raw TCP

svenwal:
This pull request is an update to pull request #233 (see also a copy of the notes below)
For the use case that Kong needs to run in raw tcp mode I have added some more code - it does work on my test environment but maybe it does not look very elegant.

1. I create within the Dockerfile (while still being root) two copies of Nginx, one with setcap run against it, one without
2. In docker-entrypoint I create a link to either the setcap ("transparent") version of Nginx or the non-setcap one depending on startup variables
3. kong prepare is now after the check for raw tcp as we need to create the Nginx link first

So when raw tcp is needed the Nginx executable has the needed permissions for raw tcp.

Oh, and I have readded the "exec" again but without the su-exec as we are not root

*** from #233 
This pull request changes the Dockerfile and entrypoint.sh in a way that the image can be run without root privileges and as an arbitrary user (anyuid). When those changes have been applied Kong can be run on a default OpenShift installation without changing any default settings regarding root user or arbitrary user.

See https://docs.openshift.com/enterprise/3.0/creating_images/guidelines.html for reference about the needed changes.

The Dockerfile changes in detail:

kong user is now member of the root group
the chown command which has been located in the entrypoint.sh until now is part of the Dockerfile (as we only have the permissions here to execute the command)
chgrp/chmod commands added so /usr/local/kong belongs to user kong's root group
USER kong added so image is run as kong from the start
The entrypoint.sh changes in detail:

chown and chgrp commands removed / relocated to Dockerfile (we are user kong in entrypoint.sh now so no permissions to do so anymore)
"exec su-exec kong" removed as we are king anyway

### DIFF
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -17,6 +17,7 @@ RUN adduser -Su 1337 -g root kong  \
 	&& cp -R /tmp/etc / \
 	&& rm -rf /tmp/etc \
 	&& apk del .build-deps \
+	&& chown -R kong /usr/local/kong \
 	&& chgrp -R 0 /usr/local/kong \
 	&& chmod -R g=u /usr/local/kong
 
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -21,6 +21,13 @@ RUN adduser -Su 1337 -g root kong  \
 	&& chgrp -R 0 /usr/local/kong \
 	&& chmod -R g=u /usr/local/kong
 
+# for transparent proxy usage enable setcup for kong
+RUN chown kong /usr/local/openresty/nginx/sbin/ \
+	&& cp /usr/local/openresty/nginx/sbin/nginx /usr/local/openresty/nginx/sbin/nginx-transparent \
+	&& cp /usr/local/openresty/nginx/sbin/nginx /usr/local/openresty/nginx/sbin/nginx-non-transparent \
+	&& rm /usr/local/openresty/nginx/sbin/nginx \
+    && setcap cap_net_raw=+ep /usr/local/openresty/nginx/sbin/nginx-transparent
+
 COPY docker-entrypoint.sh /docker-entrypoint.sh
 
 USER kong
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -16,9 +16,13 @@ RUN adduser -Su 1337 kong \
 	&& rm -rf /tmp/usr \
 	&& cp -R /tmp/etc / \
 	&& rm -rf /tmp/etc \
-	&& apk del .build-deps
+	&& apk del .build-deps \
+	&& chgrp -R 0 /usr/local/kong \
+	&& chmod -R g=u /usr/local/kong
 
 COPY docker-entrypoint.sh /docker-entrypoint.sh
+
+USER kong
 
 ENTRYPOINT ["/docker-entrypoint.sh"]
 
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -4,7 +4,7 @@ LABEL maintainer="Kong Core Team <team-core@konghq.com>"
 ENV KONG_VERSION 1.1.0rc1
 ENV KONG_SHA256 72abd186181b5ebb263c4e12db5d89cac529e2b5ca858015d44a34d560755b35
 
-RUN adduser -Su 1337 kong \
+RUN adduser -Su 1337 -g root kong  \
 	&& mkdir -p "/usr/local/kong" \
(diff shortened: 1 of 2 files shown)


hutchic:
This looks promising -- the centos container worked in a trasparent proxy setup for me the alpine image is a little unhappy though

```
cd alpine
docker build -t temp .
docker run -it temp --rm temp /bin/sh
standard_init_linux.go:207: exec user process caused "permission denied"
```

subnetmarco:
@svenwal could you also try on Alpine?

svenwal:
Tried a small in change in docker run syntax and I can execute it:

```
docker run --rm -it temp  /bin/sh
/ $ whoami
kong
```
Note that when run in this way the nginx binary is not available (as it only will be copied into position by the if statement above). We could think to have a default binary in this location and overwrite it instead of creating the link.

alpine/Dockerfile
```diff
@@ -16,10 +16,22 @@ RUN adduser -Su 1337 kong \
 	&& rm -rf /tmp/usr \
 	&& cp -R /tmp/etc / \
 	&& rm -rf /tmp/etc \
-	&& apk del .build-deps
+	&& apk del .build-deps \
+	&& chown -R kong /usr/local/kong \
+	&& chgrp -R 0 /usr/local/kong \
+	&& chmod -R g=u /usr/local/kong
+
+# for transparent proxy usage enable setcup for kong
```
james-callahan: *setcap

alpine/Dockerfile
```diff
@@ -16,10 +16,22 @@ RUN adduser -Su 1337 kong \
...
+RUN chown kong /usr/local/openresty/nginx/sbin/ \
+	&& cp /usr/local/openresty/nginx/sbin/nginx /usr/local/openresty/nginx/sbin/nginx-transparent \
+	&& cp /usr/local/openresty/nginx/sbin/nginx /usr/local/openresty/nginx/sbin/nginx-non-transparent \
+	&& rm /usr/local/openresty/nginx/sbin/nginx \
+    && setcap cap_net_raw=+ep /usr/local/openresty/nginx/sbin/nginx-transparent
```
james-callahan: This was the problem a few PRs ago: docker don't allow official images to rely on setcap.
That thread of conversation starts here: https://github.com/docker-library/official-images/pull/5180#issuecomment-447150620

alpine/docker-entrypoint.sh
```diff
@@ -12,24 +12,30 @@ if [[ "$1" == "kong" ]]; then
...
+      ln -s /usr/local/openresty/nginx/sbin/nginx-transparent /usr/local/openresty/nginx/sbin/nginx
+      kong prepare -p "$PREFIX" "$@"
+      exec /usr/local/openresty/nginx/sbin/nginx \
       -p "$PREFIX" \
       -c nginx.conf
+    else
+      ln -s /usr/local/openresty/nginx/sbin/nginx-non-transparent /usr/local/openresty/nginx/sbin/nginx
+      ls -al /usr/local/openresty/nginx/sbin/
+      kong prepare -p "$PREFIX" "$@"
+      exec /usr/local/openresty/nginx/sbin/nginx \
```
james-callahan: Can leave the `kong prepare` and the final `exec` outside of the if/else (otherwise there are indentation issues that need to be fixed)

alpine/Dockerfile
hutchic: @james-callahan is correct. 

To shorten some github threads official images *will* allow setcap inside the docker entrypoint script and inside a conditional check.

Prior art here: [consul](https://github.com/hashicorp/docker-consul/blob/663f2cb7edbd7c8e7629b520cfc1578cd43f1a8a/0.X/docker-entrypoint.sh#L91-L93) and [vault](https://github.com/hashicorp/docker-vault/blob/6ca0fa239f6a7150dfab827a0852a0031a02e028/0.X/docker-entrypoint.sh#L83-L92)

james-callahan: Also see https://github.com/docker-library/official-images/pull/5209#issuecomment-448827688 and https://github.com/docker-library/official-images/pull/5209#issuecomment-450449348

subnetmarco: Wasn't that problem already solved with https://github.com/docker-library/official-images/pull/5209?

james-callahan: @subnetmarco it was; but that apparently broke openshift. Hence this PR.

hutchic:
This should be resolved now with https://github.com/Kong/docker-kong/pull/243 being merged (haven't updated the official images yet)

PR #243, opened 2019-03-27
fix(openshift) openshift runs containers with random UID

hutchic:
If the kong container is ran with UID 0 (root) assume the desired behaviour is to `su-exec` run Kong with the Kong user (aka 1337) but if it's ran with a non 0 UID assume we want Kong to run as that user.

### DIFF
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -17,8 +17,7 @@ RUN adduser -Su 1337 kong \
 	&& cp -R /tmp/etc / \
 	&& rm -rf /tmp/etc \
 	&& apk del .build-deps \
-	&& chown -R kong /usr/local/kong \
-	&& chgrp -R 0 /usr/local/kong \
+	&& chown -R kong:0 /usr/local/kong \
 	&& chmod -R g=u /usr/local/kong
 
 COPY docker-entrypoint.sh /docker-entrypoint.sh
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -16,7 +16,10 @@ RUN adduser -Su 1337 kong \
 	&& rm -rf /tmp/usr \
 	&& cp -R /tmp/etc / \
 	&& rm -rf /tmp/etc \
-	&& apk del .build-deps
+	&& apk del .build-deps \
+	&& chown -R kong /usr/local/kong \
+	&& chgrp -R 0 /usr/local/kong \
+	&& chmod -R g=u /usr/local/kong
 
 COPY docker-entrypoint.sh /docker-entrypoint.sh
 
--- a/alpine/docker-entrypoint.sh
+++ b/alpine/docker-entrypoint.sh
@@ -14,10 +14,6 @@ if [[ "$1" == "kong" ]]; then
     shift 2
     kong prepare -p "$PREFIX" "$@"
 
-    # workaround for https://github.com/moby/moby/issues/31243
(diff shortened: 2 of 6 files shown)


alpine/Dockerfile
```diff
@@ -16,7 +16,10 @@ RUN adduser -Su 1337 kong \
 	&& rm -rf /tmp/usr \
 	&& cp -R /tmp/etc / \
 	&& rm -rf /tmp/etc \
-	&& apk del .build-deps
+	&& apk del .build-deps \
+	&& chown -R kong /usr/local/kong \
+	&& chgrp -R 0 /usr/local/kong \
```
james-callahan: No need for a separate `chgrp`, above line can be `chown -R kong:0 /usr/local/kong`.

alpine/Dockerfile
```diff
@@ -16,7 +16,10 @@ RUN adduser -Su 1337 kong \
...
+	&& chmod -R g=u /usr/local/kong
```
james-callahan: I'm curious why this isn't already the case in the archive.

alpine/docker-entrypoint.sh
```diff
@@ -13,23 +13,27 @@ if [[ "$1" == "kong" ]]; then
...
-        || has_transparent "$KONG_PROXY_LISTEN" \
-        || has_transparent "$KONG_ADMIN_LISTEN";
-    then
-      setcap cap_net_raw=+ep /usr/local/openresty/nginx/sbin/nginx
-    fi
-
-    exec su-exec kong /usr/local/openresty/nginx/sbin/nginx \
+    if [ "$(id -u)" != "0" ]; then
+      exec /usr/local/openresty/nginx/sbin/nginx \
       -p "$PREFIX" \
```
james-callahan: indentation here should be one more in.

alpine/Dockerfile
james-callahan: I think the dockerfile should gain a USER directive

hutchic: I don't think it should. Default behaviour is to run as root then su-exec to Kong. Open shift behaviour is to run as a random UID and continue as that UID

subnetmarco:
Is this PR somewhat related to https://github.com/Kong/docker-kong/pull/236 ?

hutchic:
per https://github.com/Kong/docker-kong/issues/224#issuecomment-476711301 I tried the solutions suggested https://stackoverflow.com/questions/911168/how-to-detect-if-my-shell-script-is-running-through-a-pipe for cases where `chmod o+w /proc/self/fd/1` without luck.

I was able to reproduce the `chmod: /proc/self/fd/1: Permission denied` on openshift and what I have here resolves that issue so it's possible this will work for docker swarm as well.

subnetmarco:
@hutchic can we test this patch horizontally across the board before merging? OpenShift, Vanilla K8S, Docker Swarm, and Mesos?

hutchic:
@subnetmarco
- docker-compose is automatically tested
- docker swarm hasn't worked since 1.0 so it won't regress anything
- open shift includes a automated unit test
- automated k8s test with and without mesh is on the roadmap

So overall this won't add any regressions and might fix docker swarm. I feel comfortable merging or we can wait for more automated tests / definitive conclusions

hutchic:
Tested:
- [x] docker-compose
- [x] openshift
- [x] k8s
- [x] k8s mesh

PR #252, opened 2019-04-21
Replace scale command

farwish:
scale command is deprecated

### DIFF
--- a/compose/README.md
+++ b/compose/README.md
@@ -19,7 +19,7 @@ $ docker-compose up
 To scale Kong (ie, to three instances) execute:
 
 ```shell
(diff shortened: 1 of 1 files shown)


hutchic:
I might be mistaken but isn't `docker-compose scale` valid for `docker-compose.yml` files that are version `2.1`?

If that's the case this might still be fine but we should probably have a README.md related to swarm [git://docker-kong/swarm/ here](https://github.com/Kong/docker-kong/tree/master/swarm) that includes the `>2` docker-compose commands

Thoughts?

farwish:
Valid. Thanks for your reply.

PR #259, opened 2019-05-06
fix: send QUIT instead of TERM for graceful shutdown

hbagdi:
While running in any docker environment like Docker swarm or k8s,
sending a SIGTERM to Kong will drop in-flight requests.

NGINX has decided to keep using SIGTERM because UNIX domain sockets are
not cleaned up properly while using SIGQUIT:
https://github.com/nginxinc/docker-nginx/issues/167
https://github.com/nginxinc/docker-nginx/issues/46
https://trac.nginx.org/nginx/ticket/753

OpenResty, on the other hand, uses SIGQUIT as using UNIX domain sockets
are _usually_ not used with OpenResty:
https://github.com/openresty/docker-openresty/commit/709f99f76af87009f278fdf68684c8712942d44a

### DIFF
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -26,6 +26,6 @@ ENTRYPOINT ["/docker-entrypoint.sh"]
 
 EXPOSE 8000 8443 8001 8444
 
-STOPSIGNAL SIGTERM
+STOPSIGNAL SIGQUIT
 
 CMD ["kong", "docker-start"]
--- a/centos/Dockerfile
+++ b/centos/Dockerfile
@@ -28,6 +28,6 @@ ENTRYPOINT ["/docker-entrypoint.sh"]
 
 EXPOSE 8000 8443 8001 8444
 
-STOPSIGNAL SIGTERM
+STOPSIGNAL SIGQUIT
 
 CMD ["kong", "docker-start"]
--- a/rhel/Dockerfile
+++ b/rhel/Dockerfile
@@ -29,6 +29,6 @@ ENTRYPOINT ["/docker-entrypoint.sh"]
 
 EXPOSE 8000 8443 8001 8444
 
-STOPSIGNAL SIGTERM
+STOPSIGNAL SIGQUIT
 
 CMD ["kong", "docker-start"]


thibaultcha:
We have debated using unix domains many times for various future components (e.g. sidecar plugins, redirect traffic through various ports between stream and http modules, etc...) but don't use any so far. Let's try to remember that we made this decision if we ever decide to use unix domains for any reason in the future.

Tieske:
@hbagdi this will not solve the problem of dropped requests with k8s. 

k8s is eventual consistent, so the order in which it updates its loadbalancers and the Kong servers behind it is not guaranteed. To NOT drop requests, you must first update the loadbalancers, to stop sending traffic to the to-be-stopped nodes, and only then send a `sigquit` signal. If not done this way the loadbalancers might briefly still send traffic to the node that is already shutting down and no longer accepting connections.

See the `--wait` flag: https://github.com/Kong/kong/pull/4201

hbagdi:
@Tieske:

On the Docker container side of things, once Docker sends SIGQUIT to Kong, it will stop accepting any TCP connections or new requests. Docker engine or k8s will not kill the pod but wait for it to terminate, which helps serving existing connections or in-flight requests.

> If not done this way the loadbalancers might briefly still send traffic to the node that is already shutting down and no longer accepting connections.

From the load-balancer's perspective, most load-balancers (not all) have a retry policy on  TCP failures (and may be HTTP too, but that's out of scope for this discussion). Hence, when the load balancer sees that a Kong instance is not accepting a TCP connection, it should retry another pod of Kong.

In case of our beloved k8s [1], if the load balancer is correctly configured to listen for k8s events, the load balancer should remove the terminating pod as one of it's upstream targets only after it starts terminating. 

I do agree how ideally, the load balancer should be updated first and then the node be drained (like in case of AWS ELB and EC2 instances).
The purpose here is to make Kong as friendly to all orchestration engines as possible. There are things which are outside of our control like the ELB draining nodes correctly, we should have appropriate documentation in place to minimize the hiccups that user have.

[1]: https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods

Tieske:
using the `--wait` flag you can already do this by using the `PreStop` hook. Not sure what the side effects would be but `kong quit --timeout=10 --wait=2` in pre-stop should have the usual effect, without dropping connections, even briefly while updating the LB's. Only the maximum delay for shutdown is then 12 seconds.

hbagdi:
@Tieske Please also see https://github.com/Kong/kubernetes-ingress-controller/pull/283.
`--wait` makes sense but assumes that the load balancer will drain the node before the `PreStop` is executed, which might or might not be the case.

To summarize, I think with this change, Kong is flexible enough and has sane defaults for most use cases. Which solution does the job better depends on the specific architecture of the infrastructure in which Kong is being deployed.

PR #268, opened 2019-06-18
fix(enterprise) add enterprise license option

Tieske:
Also some minor cosmetic changes and an example.


Tieske:
the 2nd commit adds a fix for a trailing comma.

PR #279, opened 2019-07-24
fixed(customize) crate missing plugins directory

bllmo:
I open this pull request because when i try to build a kong image with only a custom plugin 

```
docker build \
   --build-arg "KONG_LICENSE_DATA=$KONG_LICENSE_DATA" \
   --build-arg KONG_BASE="kong:latest" \
   --build-arg PLUGINS="kong-plugin-aws-firehose-log,kong-plugin-aws-kinesis-log" \
   --build-arg ROCKS_DIR="./rocksdir" \
   --tag "kong" .
```
I receive this error

```
...

********************************************************************************
*                          Pack newly installed rocks                          *
********************************************************************************
cd /plugins && luarocks pack kong-plugin-aws-firehose-log 0.1.0-1
[packer exec] cd /plugins && luarocks pack kong-plugin-aws-firehose-log 0.1.0-1
sh: cd: line 1: can't cd to /plugins: No such file or directory
failed packing rock: 'kong-plugin-aws-firehose-log-0.1.0-1' failed
The command '/bin/sh -c /usr/local/openresty/luajit/bin/luajit /packer.lua -- "$INJECTED_PLUGINS"' returned a non-zero code: 1
```

### DIFF
--- a/customize/packer.lua
+++ b/customize/packer.lua
@@ -201,6 +201,12 @@ local function install_plugins(plugins, lr_flag)
   end
 end
 
+local function create_plugins_directory() 
+  local ok = exec("mkdir plugins")
+  if not ok then
+    fail(("failed to create plugins dir"))
+  end
+end
 
 local function pack_rocks(rocks)
   local cmd = "cd /plugins && luarocks pack %s %s"
@@ -331,6 +337,8 @@ for plugin_name in pairs(get_plugins()) do
   end
 end
 
+header("Creare plugins directory")
+create_plugins_directory() 
 
 header("Pack newly installed rocks")
 pack_rocks(added_rocks)
@@ -403,4 +411,3 @@ stdout(script)
 
 
 header("Completed packing rocks and/or template")
-


customize/packer.lua
```diff
@@ -331,6 +337,8 @@ for plugin_name in pairs(get_plugins()) do
   end
 end
 
+header("Creare plugins directory")
```
hishamhm: There's a typo here

customize/packer.lua
```diff
@@ -201,6 +201,12 @@ local function install_plugins(plugins, lr_flag)
   end
 end
 
+local function create_plugins_directory() 
+  local ok = exec("mkdir plugins")
```
hishamhm: Better do this to stay on the safe side:

```suggestion
  local ok = exec("mkdir -p /plugins")
```

Tieske:
@bllmo I cannot reproduce this. That directory should always be there. If not, something else is wrong.

The docker file [copies the config template](https://github.com/Kong/docker-kong/blob/master/customize/Dockerfile#L19), if not specified, that template will [default to `empty_file`](https://github.com/Kong/docker-kong/blob/master/customize/Dockerfile#L10-L11). Result is; that file is always there.

The [docker `COPY` command will always create the directories](https://docs.docker.com/engine/reference/builder/#copy)

> If \<dest\> doesn’t exist, it is created along with all missing directories in its path.

Concluding; directory must be there.

What might be wrong is that you invoked the `docker build` command from a different directory (should be from the `<repo>/customize/` directory). If you do it like that, then the default value for `empty_file` does not exist, and might not be copied, and then the path might not exist...

Tieske:
closing due to no response. Please reopen if the problem reoccurs.

PR #283, opened 2019-08-02
feat(entrypoint) allow users to configure uid/gid of kong user

hbagdi:
If `KONG_UID` or `KONG_GID` environment variable is set, change the uid
and gid of the kong user (and group if it exists).

On alpine, kong group doesn't exist, so that is not allowed.

This change makes Kong compatible with Istio.

From https://istio.io/docs/setup/kubernetes/additional-setup/requirements/:

> Application UIDs: Ensure your pods do not run applications as a user
with the user ID (UID) value of 1337.


hutchic:
can't this already be accomplished by passing `--user` when running the Kong container? ie https://github.com/Kong/docker-kong/blob/master/tests.sh#L76-L85

hbagdi:
@hutchic Indeed. 🤦‍♂

PR #285, opened 2019-08-08
Updated README.md

viragumathe5:
Added information for KONG


hutchic:
I think we're OK without this addition thank you

PR #296, opened 2019-09-16
fix(openshift) allow images to run as user kong

murillopaula:
Per the OpenShift docs [1] if the Dockerfile does not have the USER
directive, it will be run as the root UID, which was the case for our
alpine and centos images. The problem is that running as root on some
OpenShift setups (e.g. Minishift) did not allow Kong to be started:

```
~/Downloads$ oc logs -f deployment/kong-rc
chmod: /proc/self/fd/1: Permission denied
chmod: /proc/self/fd/2: Permission denied
nginx: [emerg] open() "/dev/stderr" failed (13: Permission denied)
```

This fixes an OpenShift issue where images were not running as root
because no USER was specified in the alpine and centos dockerfiles.

[1]
https://docs.openshift.com/enterprise/3.2/admin_guide/manage_scc.html#enable-images-to-run-with-user-in-the-dockerfile

PR #308, opened 2019-11-26
chore(service mesh) service mesh is no longer part of kong 2.0

hutchic:
Service mesh has been migrated to https://kuma.io/

Hold merging this for Kong 2.0


alpine/Dockerfile
```diff
@@ -5,7 +5,7 @@ ENV KONG_VERSION 1.4.0
 ENV KONG_SHA256 e31080487cfb8262a839c19d5e8fdb18b7686e9e4cc594803eaa565e0b34ff7a
 
 
-RUN adduser -Su 1337 kong \
+RUN adduser -S kong \
 	&& mkdir -p "/usr/local/kong" \
 	&& apk add --no-cache --virtual .build-deps curl wget tar ca-certificates \
```
murillopaula: I know it's not a blocker for this PR nor it's relevant but can we remove packages like curl, wget, tzdata, libcap and su-exec for security purposes? I don't think these are necessary as tests were all green for the 8fb68f1 commit before: https://travis-ci.com/Kong/docker-kong/builds/129166652

I'm also going to close https://github.com/Kong/docker-kong/pull/296 since we're cherry-picking it here.

hishamhm: Isn't tzdata needed for timezones to work correctly?

PR #327, opened 2020-01-29
feat(enterprise) merge our Kong enterprise dockerfiles

hutchic:
add build-arg's so that we can build kong enterprise images from this repository.

Also included:
- got the Ubuntu CI tests passing
- adjusted the user the docker container runs kong as

Todo:
- [x] merge the kong-build-tools branch
- [x] adjust release automation script
- [x] document the user change (probably release notes of Kong)
- [x] check customize doesn't need to be fixed for the user change


Makefile
```diff
@@ -1,4 +1,4 @@
-KONG_BUILD_TOOLS?=2.0.5
```
hutchic: Merge / release this branch https://github.com/Kong/kong-build-tools/pull/213

hutchic: done

ubuntu/Dockerfile
```diff
@@ -1,16 +1,23 @@
 FROM ubuntu:xenial
 LABEL maintainer="Kong Core Team <team-core@konghq.com>"
 
-ENV KONG_VERSION 2.0.0
+# VERSION='2.0.0' community
```
Tieske: shouldn't this one also use a "downloader" to be on a par with the other ones?

Also; isn't there a way to remove the repeated downloader config out and de-duplicate it?

hutchic: There is a way to de-duplicate but it wouldn't be compatible with docker's automated build system.

Tieske: ok, but then still shouldn't this one use a downloader for consistency? so with future changes we can simply replicate the same change in every platform-file?

rhel/Dockerfile
```diff
@@ -12,16 +33,22 @@ LABEL name="Kong" \
...
+
+RUN useradd kong \
+	&& mkdir -p "/usr/local/kong" \
+	&& yum install -y /tmp/kong.rpm \
+	&& yum clean all \
+	&& rm /tmp/kong.rpm \
+	&& chown -R kong:0 /usr/local/kong \
+	&& chmod -R g=u /usr/local/kong
+
+USER kong
```
murillopaula: `USER kong` is needed because of OpenShift, right?

But having tools like curl and wget and the ability to run as root are often useful for development, and Kubernetes' lack of of a `--user` makes it difficult to use these images for testing. There are security reasons to not include tools like curl and wget and run as root but I wonder if we should include a Dockerfile.debug for development that has `USER root` and wget, curl, httpie, jq, etc. It's not a blocker for this PR but it would be useful for users wanting to use our Kong base image for development. If we don't include it, they would have to run as root and install packages in the running container. For Kubernetes it's extra work to modify build and push to the registry. However, I'm not sure if including a Dockerfile.debug is a common.

murillopaula:
@hutchic  From the checklist
```
adjust release automation script
```

Which one is it? Does it take into account the Kong Enterprise release script?

rhel/Dockerfile
hutchic: Docker development files exist in a few places including kong-build-tools, pongo, and gojira. Let's leave unifying / making an "official" Kong docker development image as a future us task.

OpenShift had challenges with previous version of Kong because it was a hard requirement that it had to run as UID 1337.

Going forward using the `USER kong` as you mentioned is for security reasons -- use the least privileged user possible

hutchic:
superseded by https://github.com/Kong/docker-kong/pull/336/

PR #336, opened 2020-02-13
chore(kong-ee) migrate the private kong-ee repository to here

hutchic:
The docker-kong-ee git repository is mostly the same as this one. This PR moves us towards merging them while keeping the flexibility of diverging when necessary.

Tested the changes to `update.sh` by using it to update CE from `2.0.0` to `2.0.1` as demonstrated in this commit https://github.com/Kong/docker-kong/pull/336/commits/04ee35993a95fc8b60d72771f926ded924926870


hutchic:
I've been toying with ways to test EE in public repositories in a safe / secure and sane way.

In the spirit of incremental improvement I'm going to merge this one shortly but there's no rush to cut Kong Enterprise over to using it right away

PR #347, opened 2020-03-10
chore(ci) remove the builder pattern from official docker images

hutchic:
docker hub image builder doesn't support the builder pattern in the way we were attempting to use it. This effectively backs that change out


alpine/Dockerfile
```diff
@@ -19,28 +18,29 @@ ARG LOCAL_KONG_PACKAGE=empty-kong.tar.gz
...
 RUN mkdir /kong; \
-	tar -C /kong -xzf /tmp/kong.tar.gz
-
-FROM alpine:3.11
-
-LABEL maintainer="Kong <support@konghq.com>"
-
-COPY --from=downloader /kong/usr/local /usr/local
-COPY --from=downloader /kong/etc /etc
+	tar -C /kong -xzf /tmp/kong.tar.gz && \
```
gszr: Obtaining the package in a separate step (the copy at the top) will cause the final image to have one unneeded extra layer, increasing the image size, compared to old versions of the Dockerfile.

alpine/Dockerfile
```diff
@@ -19,28 +18,29 @@ ARG LOCAL_KONG_PACKAGE=empty-kong.tar.gz
...
     && mkdir -p "/usr/local/kong" \
 	&& chown -R kong:0 /usr/local/kong \
+	&& chown kong:0 /usr/local/bin/kong \
 	&& chmod -R g=u /usr/local/kong
 
-USER kong
-
 COPY docker-entrypoint.sh /docker-entrypoint.sh
 
+RUN chmod 777 /docker-entrypoint.sh
```
gszr: Is this 777 required?

hutchic: It was on my local but I think it's because something has (mis)configured security setup. I'll nuke it and if it passes CI consider it OK :)

alpine/Dockerfile
hutchic: The `COPY` command is copying an empty file which results in a 0 byte noop. It adds a layer but I don't think there's anyways to `IF` a `COPY` command currently

PR #350, opened 2020-03-19
chore(admin) restrict the admin port to 127

hutchic:


tests.sh
```diff
@@ -29,7 +29,7 @@ done
 
 sleep 20
 curl -I localhost:8001 | grep 'Server: openresty'
-
+sed -i -e 's/127.0.0.1://g' docker-compose.yml
```
kidd: as a suggestion, maybe using an env var that makes explicit inthe docker-compose.yml file that things might change from the outside.

PR #352, opened 2020-03-23
refactor(test) use explicit env var instead of sed

kidd:
This PR removes a sed in favor of using an env var. Would 0.0.0.0 work as a default? or better just leaving it empty?

Not sure if we could even look for some more commonality by reusing KONG_ADMIN_LISTEN environment var, but one step at a time


Tieske:
since this is the admin api it should default to 127.0.0.1 for security reasons, not accidentally opening a backdoor

PR #372, opened 2020-05-27
added tini for the dockerfiles

srinidhis05:
The fix to use dumb init / tini as a lite init system was added in [https://github.com/Kong/docker-kong/pull/48](https://github.com/Kong/docker-kong/pull/48), but seem to have missed in the docker files that are present currently. 
The PR is to add tini as a init system for the dockerfiles .
Ref: 
[docker-and-the-pid-1-zombie-reaping-problem]( https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/.)

### DIFF
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -36,9 +36,14 @@ RUN set -ex; \
 
 COPY docker-entrypoint.sh /docker-entrypoint.sh
 
+# Add Tini
+ENV TINI_VERSION v0.10.0
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
+RUN chmod +x /tini
+
 USER kong
 
-ENTRYPOINT ["/docker-entrypoint.sh"]
+ENTRYPOINT ["/tini", "--", "/docker-entrypoint.sh"]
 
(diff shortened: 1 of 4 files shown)


Abhishekvrshny:
cc: @hutchic

Abhishekvrshny:
@hutchic The `go-pluginserver` can be a different image, which can be deployed as a sidecar, daemon set or as a standalone deployment. The `go-pluginserver` can again be spawned under tini in this case as `systemd` would not be supported in a normal docker container.

hutchic:
Agreed it can be done that route, just discussing if we will go that route. If we did I don't think we'd go include `tini`.

Are there strong reasons I'm not seeing to use `tini`? My impression is it's a light process management tool to avoid zombies. So long as one can validate kong's healthiness / readiness process management kind of falls on the orchestration layer and arguably none of it should be done within the docker image itself

Tieske:
@hutchic any update on this?

hutchic:
I don't believe it benefits us / isn't widely asked for. I think we can close and folks that wish to use tini can build a container `FROM kong` or if demand does increase we can make the change

Tieske:
ok, let's close for now.

PR #383, opened 2020-07-03
fix(ee) set EE ASSET

kidd:


alpine/Dockerfile
```diff
@@ -33,7 +33,9 @@ RUN set -ex; \
 	chown kong:0 /usr/local/bin/kong && \
 	chmod -R g=u /usr/local/kong && \
 	rm -rf /tmp/kong.tar.gz && \
-	kong version
+  if [ "$ASSET" = "ce" ] ; then \
+    kong version ; \
+  fi;
```
hutchic, line 38: is there a different command EE allows that doesn't require a license? This is a nice sanity check to have but I understand if it's not possible.

We could maybe `test -f /usr/local/bin/kong`?

alpine/Dockerfile
```diff
@@ -33,7 +33,9 @@ RUN set -ex; \
 	chown kong:0 /usr/local/bin/kong && \
```
kidd, line 33: Answering to your question down there, this line will fail if the file doesn't exist, so it's already a `-f /usr/local/bin/kong`

Issue #379, opened 2020-06-29
2.1 beta 1:  missing database for connect


luispabon:
```
nginx: [error] init_by_lua error: /usr/local/share/lua/5.1/pgmoon/init.lua:550: missing database for connect
stack traceback:
	[C]: in function 'assert'
	/usr/local/share/lua/5.1/pgmoon/init.lua:550: in function 'send_startup_message'
	/usr/local/share/lua/5.1/pgmoon/init.lua:207: in function 'connect'
	.../share/lua/5.1/kong/db/strategies/postgres/connector.lua:211: in function 'connect'
	.../share/lua/5.1/kong/db/strategies/postgres/connector.lua:527: in function 'query'
	.../share/lua/5.1/kong/db/strategies/postgres/connector.lua:279: in function 'init'
	/usr/local/share/lua/5.1/kong/db/init.lua:139: in function 'init_connector'
	/usr/local/share/lua/5.1/kong/init.lua:419: in function 'init'
	init_by_lua:3: in main chunk
```

This is happening with the normal and alpine versions of the image. Weirdly enough, it does not happen with the migrations:

kong migrations up:
```
migrating core on database 'kong'...
core migrated up to: 009_200_to_210 (pending)
migrating rate-limiting on database 'kong'...
rate-limiting migrated up to: 004_200_to_210 (executed)
migrating hmac-auth on database 'kong'...
hmac-auth migrated up to: 003_200_to_210 (pending)
migrating oauth2 on database 'kong'...
oauth2 migrated up to: 004_200_to_210 (pending)
migrating ip-restriction on database 'kong'...
ip-restriction migrated up to: 001_200_to_210 (pending)
migrating jwt on database 'kong'...
jwt migrated up to: 003_200_to_210 (pending)
migrating basic-auth on database 'kong'...
basic-auth migrated up to: 003_200_to_210 (pending)
migrating key-auth on database 'kong'...
key-auth migrated up to: 003_200_to_210 (pending)
migrating acl on database 'kong'...
acl migrated up to: 003_200_to_210 (pending)
migrating bot-detection on database 'kong'...
bot-detection migrated up to: 001_200_to_210 (pending)
10 migrations processed
1 executed
9 pending
```

kong migrations finish
```
migrating core on database 'kong'...
core migrated up to: 009_200_to_210 (executed)
migrating hmac-auth on database 'kong'...
hmac-auth migrated up to: 003_200_to_210 (executed)
migrating oauth2 on database 'kong'...
[...]

PR #386, opened 2020-07-09
fix(docker) do not set empty vars

eskerda:
file_env sets empty envs when is run. This PR adds a safeguard for that case. If `<smth>_FILE` is not set the function does not need to run.


eskerda:
@hutchic do you see any downside on this change?

hutchic:
LGTM

hutchic:
Updating official images with this will fix https://github.com/Kong/docker-kong/issues/379

PR #388, opened 2020-07-10
fix(dependencies) remove uneeded dev dependencies

hutchic:
during the 2.0.5 release the docker team questioned the need for these `-dev` packages. After some sleuthing / testing we've determined they're not required

### DIFF
--- a/ubuntu/Dockerfile
+++ b/ubuntu/Dockerfile
@@ -18,9 +18,7 @@ RUN set -ex; \
         && apt-get purge -y curl; \
     fi; \
     apt-get update \
-    && apt-get install -y --no-install-recommends perl unzip git \
-    && { apt-get install -y --no-install-recommends zlibc || true; } \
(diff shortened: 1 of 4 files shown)


ubuntu/Dockerfile
```diff
@@ -20,7 +20,6 @@ RUN set -ex; \
     apt-get update \
     && apt-get install -y --no-install-recommends perl unzip git \
     && { apt-get install -y --no-install-recommends zlibc || true; } \
-    && { apt-get install -y --no-install-recommends zlib1g-dev || true; } \
```
gszr, line 23: need to replace both with `zlibc1g`

hutchic:
https://github.com/hamishforbes/lua-ffi-zlib/blob/master/lib/ffi-zlib.lua#L98 tries to load the `zlib.so` via the symlink that's only included in the `-dev` packages

PR #389, opened 2020-07-10
fix(ubuntu) do not run kong version

eskerda:


ubuntu/Dockerfile
```diff
@@ -29,14 +29,14 @@ RUN set -ex; \
 	&& chown -R kong:0 /usr/local/kong \
 	&& chown kong:0 /usr/local/bin/kong \
 	&& chmod -R g=u /usr/local/kong \
-	&& kong version
+    && if [ "$ASSET" = "ce" ] ; then \
+	kong version ; \
+    fi;
```
hutchic, line 34: spacing looks slightly inconsistent

eskerda: Used spaces, I think now it's consistent with the `if` on top. Is that what you meant, or you meant regarding the other Dockerfiles?

PR #391, opened 2020-07-10
chore(bintray-release) add beta destinations

kidd:


bintray-release.sh
```diff
@@ -115,8 +115,19 @@ if [[ "$KONG_RELEASE" == "internal-preview" ]] ; then
     INTERNAL_PREVIEW=true
 fi
 
+if [[ "$KONG_RELEASE" == beta* ]] ; then
+    BETA=true
```
eskerda, line 119: Where is this env used?

kidd: nope,... removing it :+1:

bintray-release.sh
```diff
@@ -115,8 +115,18 @@ if [[ "$KONG_RELEASE" == "internal-preview" ]] ; then
     INTERNAL_PREVIEW=true
 fi
 
+if [[ "$KONG_RELEASE" == beta* ]] ; then
+    TAG_LATEST=
```
eskerda, line 119: Reminder for a future plan, these should be over loadable instead (or plus) of being a flag, so this kind of behavior can be tamed on run.

PR #409, opened 2020-08-27
chore(dependencies) infer dependencies from the packages

hutchic:
fpm marks the zlib dependencies therefore it shouldn't be necessary to duplicate that logic in our Dockerfiles

### DIFF
--- a/centos/Dockerfile
+++ b/centos/Dockerfile
@@ -11,19 +11,19 @@ COPY kong.rpm /tmp/kong.rpm
 ARG KONG_VERSION=2.1.3
 ENV KONG_VERSION $KONG_VERSION
 
-ARG KONG_SHA256="be97835859ca05de3db5144bb59c8bdd4e46bd7a65fdd5f7ec500cb445eb62e3"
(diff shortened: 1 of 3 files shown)


centos/Dockerfile
```diff
@@ -11,19 +11,19 @@ COPY kong.rpm /tmp/kong.rpm
 ARG KONG_VERSION=2.1.3
 ENV KONG_VERSION $KONG_VERSION
 
-ARG KONG_SHA256="be97835859ca05de3db5144bb59c8bdd4e46bd7a65fdd5f7ec500cb445eb62e3"
+ARG KONG_SHA256="c39e596ac3aac771377fc67eb4fb151216daeba6e8b1595c670e8c571b5cd3a0"
```
kidd, line 14: are we also bumping a version?  this might have slipped in

kidd (commented):
Didn't try it manually, but I assume it's ok.

kidd (approved):

centos/Dockerfile
hutchic: we did a re-release to change the dependencies. The asset is the same but the signature changed

PR #413, opened 2020-09-16
chore(*) update install docs if dockerfile install changes

murillopaula:


murillopaula:
Question: should have a debian dockerfile? Right now we're relying on the ubuntu dockerfile to test debian releases looks like: https://github.com/Kong/kong-build-tools/blob/master/test/build_container.sh#L8-L10

murillopaula:
Question 2: should we be running CI tests against non-docker installs too? Right now we run tests against docker images only but docker images aren't always equivalent to distributions. (I know this is way out of scope of this PR, but just asking)

gszr:
> Question: should have a debian dockerfile? Right now we're relying on the ubuntu dockerfile to test debian releases looks like: https://github.com/Kong/kong-build-tools/blob/master/test/build_container.sh#L8-L10

I wouldn't personally oppose to this, but would be nice to see if there's been any asks from the community / customers.

gszr:
> Question 2: should we be running CI tests against non-docker installs too? Right now we run tests against docker images only but docker images aren't always equivalent to distributions. (I know this is way out of scope of this PR, but just asking)
> 
> edit: maybe we could do this via the jenkins cloud plugin, would it be worth the effort?

Unsure I understood this correctly, but the packages are smoke-tested separately, in kong-build-tools.

gszr (approved):
LGTM! Needs a rebase though

murillopaula:
> I wouldn't personally oppose to this, but would be nice to see if there's been any asks from the community / customers.

No demand yet AFAIK. And no issues reported so far.

murillopaula:
> Unsure I understood this correctly, but the packages are smoke-tested separately, in kong-build-tools. Mind clarifying?

They are smoke-tested in a docker image, my idea was to run smoke tests not in docker images, but in actual distributions.

PR #420, opened 2020-10-01
chore(tests) use the kong source that matches the version we built

hutchic:
Edit to add: there's no historic case defaulting to next would not have worked.

We clone the kong source so we have access to the `.requirements` file.

The requirements file is needed by kong-build-tools to double check the right things got built. However, running the  [01-package test suite](https://github.com/Kong/kong-build-tools/tree/master/test/tests/01-package) on docker-kong CI is largely unecessary (with the exception of the version check as it has been beneficial in the past) as any asset couldn't have been built without already going through the tests.

I think this initially got pinned to `2.1.0-beta.1` because the `$version_given` variable is incorrect for our prereleases.

Future improvements:
- only run the [02-api test suite](https://github.com/Kong/kong-build-tools/tree/master/test/tests/02-api) against docker-kong
- logic to handle pre release tag format. In the absence of a tag default to `next` might be OK
- possibly merge kbt and docker-kong. They're becoming rather intertwined

### DIFF
--- a/tests/image.test.sh
+++ b/tests/image.test.sh
@@ -110,7 +110,7 @@ function run_test {
 
   git clone https://github.com/Kong/kong.git || true
   pushd kong
-  git checkout 2.1.0-beta.1
+  git checkout $version_given
   popd
 
   pushd kong-build-tools
--- a/tests/image.test.sh
+++ b/tests/image.test.sh
@@ -110,7 +110,7 @@ function run_test {
 
   git clone https://github.com/Kong/kong.git || true
   pushd kong
-  git checkout $version_given
+  git checkout $version_given || git checkout next
   popd
 
   pushd kong-build-tools
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-KONG_BUILD_TOOLS?=4.8.0
+KONG_BUILD_TOOLS?=4.10.1
 BASE?=centos
 
 build:


Tieske:
instead of passing the version, can't you use labels;

* https://docs.docker.com/config/labels-custom-metadata/
* https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys

similar to this: https://github.com/Kong/kong-ee/pull/1623/files

not sure though it even makes sense? is this based on images?

hutchic:
The version is snagged from the Dockerfile's themselves ie https://github.com/Kong/docker-kong/blob/master/alpine/Dockerfile#L12

The intent is:
The Dockerfile says to build x.y.z
OK make sure the `.requirements` file for `x.y.z` matches what we've actually ended up with

Tieske:
you cannot trust this. It is the exact same problem that Pongo has, it needs the matching git commit to fetch the Kong sources. Hence why the Kong-EE nightly image is tagged with the commit hash in the PR linked above.

Pongo cannot test CE nightlies because this tag is missing from those images. So maybe adding the tag would solve both problems?

hutchic:
This I'm pretty sure is a different problem (or I'm confused lol). In my case I can trust that the Dockerfile `KONG_VERSION` has an existing github tag which is true 90% of the time. For the exceptions I think defaulting back to `next` will be fine

To review the sequence is:
- kong/kbt builds a kong asset and sends it to bintray
- someone (a robot) updates docker-kong Dockerfile with the new version
- docker-kong CI pulls the asset from bintray and builds

It's at this point that we need to test the image we built is correct and to do that we need the Kong source. The exception is that prerelease `KONG_VERSION` doesn't equal the github tag ie `ARG KONG_VERSION=2.1.0beta.1` vs `2.1.0-beta.1`

tests/image.test.sh
```diff
@@ -110,7 +110,7 @@ function run_test {
 
   git clone https://github.com/Kong/kong.git || true
   pushd kong
-  git checkout $version_given
+  git checkout $version_given || git checkout next
```
Tieske, line 113: if we do not have a version, shouldn't we use a commit-id? if another PR gets merged before this runs, it will checkout the wrong code

Tieske:
for tagged versions the problem doesn't exist. Only for untagged ones, like the RC's or nightlies, and for those the problem is the exact same.

I'd prefer we fix this problem, because it would also allow Pongo to test against nightly-CE images. For EE the ability already works, but not for CE.

tests/image.test.sh
hutchic: where would the commit id come from? The docker-kong CI runs fully isolated therefore it only has availbe to it the information that exists in git://kong/docker-kong and the bintray asset itself

Tieske: Can't we pass info along with the bintray asset? a metadata file or something similar?

hutchic: Possibly. I've long thought Kong should have the equivalent of a `package.lock` on the file system that indicates Kong version, git commit, etc.

In lieu of that for official releases everything git://docker-kong has available is what we have on bintray ie https://bintray.com/kong/kong-deb/debian/2.1.4#files

Unofficial builds we have more flexibility ....

Tieske: I think that would  do, though it would be nice to have the same info available on all builds.

Anything you need from me to help implement this?

hutchic: I believe this should accomplish revision for our unofficial images

https://github.com/Kong/kong-build-tools/pull/342

PR #421, opened 2020-10-02
chore(*) protect against error with the useradd instruction

murillopaula:
The 'kong' user should now be already created on package creation [1].

The `chown` and `chmod` commands are unchanged because these are
specific commands needed to run Kong on OpenShift [2].

[1] https://github.com/Kong/kong-build-tools/pull/339
[2] https://github.com/Kong/docker-kong/pull/243/


hutchic:
I think this is fine and will pass CI after we roll out a GA package that handles user creation. Tentative 👍

PR #428, opened 2020-10-20
fix(tests) dont double test our packages

hutchic:
kbt tests run the 01 packaging tests then put the assets onto bintray
docker-kong Dockerfile's pull said assets from bintray

let's not re-run the KBT packaging tests. The KBT api tests still has use so going to keep it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,10 @@ services:
 
 env:
   matrix:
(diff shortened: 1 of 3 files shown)


Tieske (approved):
lgtm

murillopaula (approved):

PR #423, opened 2020-10-06
chore(ownership) after-install now handles file / directory ownership

hutchic:
do not merge until after 2.2.0

PR #432, opened 2020-10-28
fix(*) command grouping

gszr:


murillopaula:
no need to, lets merge https://github.com/Kong/docker-kong/pull/423

hutchic:
@murillopaula #423 is failing though 😞 

I think we could do this to get the release out and then circle back as to what's wrong the #423?

hutchic (approved):

PR #433, opened 2020-10-28
feat(customize) allow rocks with C-code (self contained)

Tieske:
Besides customizing Kong images by adding pure-Lua plugins in a docker image, this allows to also add plugins/rocks that have c-code. Luarocks makes it sooooo easy, just pack'em and ship'em.

Just rocks that bind to an existing library cannot be added yet, since we have no means of first adding that library yet.


hutchic (approved):
the tests look sensible enough so 👍

PR #438, opened 2020-11-03
fix(rhel) parameterize rhel dockerfile

gszr:


hutchic (approved):
LGTM

As a reminder we cannot parameterize alpine / ubuntu as the official docker image builder can't handle parameters

murillopaula (approved):
lgtm! need to see how CI will behave when releasing

PR #440, opened 2020-12-04
fix(update) ensure alpine and rhel update correctly

gszr:
Also update kbt version in makefile to match kong's -- and add automation to `update.sh`

### DIFF
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -12,11 +12,17 @@ COPY kong.tar.gz /tmp/kong.tar.gz
 ARG KONG_VERSION=2.2.1
 ENV KONG_VERSION $KONG_VERSION
 
+ARG KONG_AMD64_SHA="e05c7075ae263bb7ef84b0370bf4e31f5246a41d910e5d4336216ae95b654af1"
+ENV KONG_AMD64_SHA $KONG_AMD64_SHA
+
+ARG KONG_ARM64_SHA="c7b2c0e1f47e3009e55b492344bd1ff689a41eca91c21f2c3f57f9172932c96d"
+ENV KONG_ARM64_SHA $KONG_ARM64_SHA
+
 RUN set -eux; \
 	arch="$(apk --print-arch)"; \
 	case "${arch}" in \
-		x86_64) arch='amd64'; KONG_SHA256='e05c7075ae263bb7ef84b0370bf4e31f5246a41d910e5d4336216ae95b654af1' ;; \
-		aarch64) arch='arm64'; KONG_SHA256='c7b2c0e1f47e3009e55b492344bd1ff689a41eca91c21f2c3f57f9172932c96d' ;; \
+		x86_64) arch='amd64'; KONG_SHA256=$KONG_AMD64_SHA ;; \
+		aarch64) arch='arm64'; KONG_SHA256=$KONG_ARM64_SHA ;; \
 	esac; \
     if [ "$ASSET" = "ce" ] ; then \
         apk add --no-cache --virtual .build-deps curl wget tar ca-certificates && \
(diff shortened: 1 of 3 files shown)


alpine/Dockerfile
```diff
@@ -12,11 +12,17 @@ COPY kong.tar.gz /tmp/kong.tar.gz
 ARG KONG_VERSION=2.2.1
 ENV KONG_VERSION $KONG_VERSION
 
+ARG KONG_AMD64_SHA="e05c7075ae263bb7ef84b0370bf4e31f5246a41d910e5d4336216ae95b654af1"
+ENV KONG_AMD64_SHA $KONG_AMD64_SHA
+
+ARG KONG_ARM64_SHA="c7b2c0e1f47e3009e55b492344bd1ff689a41eca91c21f2c3f57f9172932c96d"
+ENV KONG_ARM64_SHA $KONG_ARM64_SHA
```
hutchic, line 19: I think this is unsupported in the official docker build system

update.sh
```diff
@@ -23,12 +23,22 @@ function die() {
 
 hub --version &> /dev/null || die "hub is not in PATH. Get it from https://github.com/github/hub"
 
+kbt_in_kong_v=$(curl -sL https://raw.githubusercontent.com/Kong/kong/$version/Makefile | grep 'KONG_BUILD_TOOLS\s*?=' | awk -F"=" '{print $2}' | tr -d "'[:space:]")
+sed -i -e 's/KONG_BUILD_TOOLS?=.*/KONG_BUILD_TOOLS?='$kbt_in_kong_v'/g' Makefile
```
hutchic, line 27: clever 👍

update.sh
```diff
@@ -23,12 +23,22 @@ function die() {
...
-   sed -i -e 's/ARG KONG_SHA256=.*/ARG KONG_SHA256=\"'$new_sha'\"/g' Dockerfile
+
+   sed -i -e 's/ARG KONG_AMD64_SHA=.*/ARG KONG_AMD64_SHA=\"'$new_sha'\"/g' Dockerfile
+   sed -i -e 's/ARG KONG_VERSION=.*/ARG KONG_VERSION='$version'/g' Dockerfile
+
+   url=$(grep bintray.com Dockerfile | awk -F" " '{print $3}' | sed 's/\"//g' | sed 's/$KONG_VERSION/'$version'/g' | sed 's/$arch/'arm64'/g')
+   curl -fL $url -o /tmp/kong
+   new_sha=$(sha256sum /tmp/kong | cut -b1-64)
+
+   sed -i -e 's/ARG KONG_ARM64_SHA=.*/ARG KONG_ARM64_SHA=\"'$new_sha'\"/g' Dockerfile
```
hutchic, line 41: full disclosure: this looks correct but I didn't actually test it 👍

hutchic (changes requested):
we've had challenges in the past with using `ARG thing=default` with the official docker repository

alpine/Dockerfile
gszr: @tianon Apologies to bug, but could you comment -- is this something unsupported by the Docker Library build system?

hutchic: could be I'm thinking of
> Unfortunately our build systems cannot handle variables in COPY or FROM statements.

trying to locate a past PR where we had thing setup in a similar fashion to see why we didn't go forward with it

tianon: IMO it's a bit "extra", but it's fine (and is done for `KONG_VERSION` just above this).

gszr: Thanks for commenting, @tianon!

hutchic (approved):

PR #447, opened 2021-01-18
fix(rhel) builder does not support vars

gszr:
This change reverts the usage of variables in the RHEL Dockerfile, as this is a feature not supported by the RedHat container catalog builder. A sister PR will follow in kong-build-tools.


gszr:
this is no longer needed, since rhel images are built by travis, not the container catalog builder

Issue #468, opened 2021-06-01
[RFE] Add support for base64 encoded KONG_LICENSE_DATA string


usrbinkat:
Issue: 
When writing IaC in a mix of HCL and json, passing license into container via environment variable can become problematic due to complexities of nested json.

Proposal:
Add conditional base64 encoded license ternary check to container entrypoint to decode license data string if found.

mflendrich:
@usrbinkat why was that added to the Team Kubernetes board? I assume it was a misclick - please correct me otherwise.

usrbinkat:
I do not believe I've been in the Team Kubernetes board so I am unsure.

Closing for now

PR #469, opened 2021-06-01
add support for base64 encoded kong license data

usrbinkat:
Proposal for adding base64 encoded kong license data support #468

### DIFF
--- a/alpine/docker-entrypoint.sh
+++ b/alpine/docker-entrypoint.sh
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 set -Eeo pipefail
 
+# Enables support for base64 encoded KONG_LICENSE_DATA_B64 string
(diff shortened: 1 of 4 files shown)


gszr (commented):
This looks good to me! Thanks @usrbinkat.

Could you please rebase onto latest master so tests run?

Tieske:
Should we be adding it here? in the docker images? Wouldn't it make more sense to do this in Kong itself?

Tieske:
had a quick look; this function https://github.com/Kong/kong-ee/blob/next/2.4.x.x/kong/enterprise_edition/license_helpers.lua#L41-L92 gets the license data. If we just try b64-decode before returning it, return the decoded result, or if the decode fails, return the input. That should be backward compatible, and work across all distributions, not just Docker.

bungle:
This adds full base64 and base64url support to Kong directly:
https://github.com/Kong/kong/pull/9379

which is about to be merged here:
https://github.com/Kong/kong/pull/9253

usrbinkat:
Closing in favor of https://github.com/Kong/kong/pull/9379

PR #477, opened 2021-06-21
Dynamically get the binary file

khiemdoan:


gszr:
Hey @khiemdoan,

Thanks for the PR. However, as the dockerfile itself is for Ubuntu, I don't see a gain in dynamically setting this. Closing this PR for now.

khiemdoan:
Hi @gszr,

This commit helps us use all Debian and Ubuntu bases. We can switch to Debian easily.

PR #478, opened 2021-06-24
feat(*) make core executables available

Tieske:
This allows the Kong container itself to be easier used as a
build container for Kong related tasks.

Adds tests for availability in the path of; luarocks, resty,
nginx, lua, luajit.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+root = true
+
+[*]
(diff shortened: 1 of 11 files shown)


Tieske:
also fixed some indentation, so probably easiest to review with "whitespace changes" disabled.

gszr (approved):

PR #480, opened 2021-06-30
fix(alpine) create a system kong group

murillopaula:
https://docs.konghq.com/enterprise/2.4.x/property-reference/#nginx_user

> Note: If the kong user and the kong group are not available, the default user and group credentials will be nobody nobody.

We must create a kong group so that the default `nginx_user` is used. This is already the case in Docker CentOS, for example, but not in Alpine.

Alpine:
```
$ id
uid=100(kong) gid=65533(nogroup) groups=65533(nogroup)
```

CentOS:
```
$ id
uid=1000(kong) gid=1000(kong) groups=1000(kong)
```

### DIFF
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -38,6 +38,7 @@ RUN set -eux; \
     && rm -rf /kong \
     && apk add --no-cache libstdc++ libgcc openssl pcre perl tzdata libcap zip bash zlib zlib-dev git ca-certificates \
     && adduser -S kong \
+    && addgroup -S kong \
     && mkdir -p "/usr/local/kong" \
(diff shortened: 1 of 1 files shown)


Tieske:
I don't know enough about unix permissions to validate this. But I am wondering whether it should get a test?

hutchic (approved):
looks sensible enough

murillopaula:
this test should be enough https://github.com/Kong/docker-kong/blob/master/tests/01-image.test.sh#L69-L83

murillopaula:
@Tieske do you approve?

Tieske (approved):

PR #482, opened 2021-07-13
Release: 2.5.0

gszr:


update.sh
```diff
@@ -51,13 +51,15 @@ fi
 
 pushd alpine
    url=$(get_url Dockerfile amd64)
+   echo $url
```
locao, line 54: Is this (and the others below it) a debug leftover?

gszr: I think this is useful to leave in - re-running to get that piece of info can be avoided.

locao (approved):

PR #484, opened 2021-07-19
CPUs should be string, not int

shaisachs:
Fixes this issue when trying `docker-compose up`:
`* 'Deploy.Resources.Limits.cpus' expected type 'string', got unconvertible type 'int', value: '2'`


Tieske (approved):
Thx!

Issue #488, opened 2021-08-03
KONG_PASSWORD_FILE support in Docker entry point


ludovic-pourrat:
We need to secure the the Kong master admin password and our selected option is to lookup a `_FILE` environment variable from the Docker entry point script.

This would imply to add the following line to https://github.com/Kong/docker-kong/blob/master/alpine/docker-entrypoint.sh right atfer the other file_env evaluatiuons.

`file_env KONG_PASSWORD`

Would this make sense ? If so a PR will be issued accordingly.

Thanks.

gszr:
Hey @ludovic-pourrat,

That does make sense. Please feel free to open a PR for that and we will get it reviewed.

Thanks for the proposal!

Tieske:
wondering whether we should add more properties in this scheme. Or maybe even make all of them configurable through secrets.

If we parse `/usr/local/share/lua/5.1/kong/templates/kong_defaults.lua` we could loop over all of them.

Tieske:
ping @PaulFischer-Kong FYI, more secrets....

gszr:
In the interest of iterative improvement, I propose we proceed with the @ludovic-pourrat's proposal -- adding `KONG_PASSWORD` to the mix -- and circle back with other secrets. What do you think @Tieske? What other secrets would you propose adding in here?

gszr:
A slight concern I have: `KONG_PASSWORD` would apply in the context of Kong Gateway (Enterprise), correct? If so, we should hide or emit an alert if this is used in OSS.

Tieske:
> A slight concern I have: `KONG_PASSWORD` would apply in the context of Kong Gateway (Enterprise), correct? If so, we should hide or emit an alert if this is used in OSS.

Why? it will not do what it was designed or, and hence that should suffice towards the user. My 2cts.

Tieske:
> What other secrets would you propose adding in here?

adding individual secrets (essentially maintaining a list in the script) requires overhead/maintenance over time. If you loop over all properties, then we no longer need to maintain a list.

Tieske:
here's the implementation: https://github.com/Kong/docker-kong/pull/490

gszr:
Hey @ludovic-pourrat, @Tieske's proposal #490 was merged. You can try it out by building a local image -- or wait til our upcoming 2.5.1 is released (should be coming in the next few days). Closing this for now. Thanks for the suggestion!

PR #490, opened 2021-08-10
feat(secrets) optional read all secrets from files

Tieske:
Instead of reading only specified values from files as secrets, this change will attempt to read all settings from a file. So any additional secrets besides the ones now in the list will also be configurable by file.

Adds `KONG_PASSWORD` specifically since that is not a regular Kong option.

Implements #488 

The way the file is parsed can probably be improved, open to any suggestions @gszr @hutchic


Tieske:
Also; only updated Alpine, if deemed ok, I'll apply the same changes to the other images

hutchic:
the tests pass so I trust it's backward compatible 👍

hutchic (approved):

gszr (commented):
Are there tests for reading secrets from files? Could we add a test case? Otherwise, LGTM.

hutchic:
@gszr there is one currently. The docker-compose stores the postgresql password in a file

gszr (approved):

PR #495, opened 2021-08-31
chore(dep): alpine version bump

hutchic:


gszr (approved):

Tieske:
@gszr wasn't the 3.14 release the one that caused issues? when the nightly ce images was build against `latest`?

gszr:
Great catch @Tieske.

Could you refresh my mind on what the issue was? luarocks-related IIRC?

I remember having it on the core-team repo (core-trello tool); just ran a quick test and that one is working fine.

gszr:
(Let's hold on merging for now)

Tieske:
here's some info: https://github.com/Kong/kong/pull/7529

hutchic:
can we add a naive pongo test? We do need this upgrade at some point 3.13 has a vulnerability

Tieske:
from this comment; https://github.com/alpinelinux/docker-alpine/issues/146#issuecomment-869595012 I get that using docker `20.10.0` should fix it.

Tieske:
problem with Pongo tests is that it requires Pongo to be updated to support that version. Current docker files are pinned at latest release Kong version. That would break upon makin g a new release, because Pongo only gets updated after the release.

Best bet might be to make sure our nightly images are build using these official docker files (which they currently are not) because then the nightly-ce based tests for plugins would fail elsewhere and show the problem.

hutchic:
@Tieske 
> make sure our nightly images are build using these official docker files (which they currently are not)

Not to be pedantic but they are the official docker files but we needed allow for the ability to manipulate the base image

kong/Jenkinsfile (release per commit) 
-> kong/Makefile (release) 
-> kong-build-tools/Makefile (release-kong) 
-> kong-build-tools/Makefile (test)
-> kong-build-tools/Makefile (build-test-container)
-> kong-build-tools/test/build_container.sh
-> clones git://kong/docker-kong
-> Dockerfile with the base image changed via sed or arguments

https://github.com/Kong/kong-build-tools/blob/f2280a8ed94d7c844ee33d0c798a077c7db5d3b1/test/build_container.sh#L42

The reason we want to manipulate the base image is we use Docker to test our release assets. For example when we build a centos 6 asset we want to test that asset in a centos:6 container. The reason we use sed instead of docker args is the docker official image builder is obnoxious and won't support base images based on arg's.

If I've got the logic train correct .... I think we could ensure the dailies are identical to the offical by passing the `OFFICIAL_RELEASE` env to `build_container.sh` then elseif only sed if it's not an official release?

hutchic:
> problem with Pongo tests is that it requires Pongo to be updated to support that version

I was advocating somewhere in our pre-release CI there should be a test that does "does this image" work with pongo. docker-kong seems like the sensible place since this is the repository where changes made here might break pongo later on?

Tieske:
Discussed with @hutchic and @mikefero 

- we consider a December 2020 docker release a fair fix to the problem, hence we'll upgrade to `3.14`
- nightlies go back to `alpine:latest`, to break ahead of anything official if there are any issues
- we need to add a comment to the changelogs for future (patch) releases (also informs support)

gszr:
I updated https://github.com/Kong/kong/pull/7820 to implement the above -- using latest in daily and 3.14 in finals.

Tieske:
then I think we're done here. @mikefero and @HCloward will take care of the changelogs

dropping the "do not merge" label, and approving

Tieske (approved):

Tieske:
@hutchic please merge when you see fit

PR #497, opened 2021-09-02
tests(swarm): remove the docker-swam tests but keep the upgrade capability test

hutchic:
docker-compose makes it challenging to have a single docker-compose file that fullfills docker-compose and docker swarm deployment capability. If we want to continue to test against docker swarm we should make a separate compose file. 

At this time I think we're OK to remove the tests that validate our compose file against docker swarm unless swarm gets a future boost in popularity I doubt we reimplement this

recap:

docker-compose file specification is a way to describe how docker container relate to each other
docker-compose cli is a way to run a docker-compose file on an orchestration system including a single docker daemon
docker swarm is an orchestration system which the docker-compose cli is compatible with and docker-compose is partially compatible with


kikito (approved):
LGTM. The "partially compatible" part is also "poorly documented", to boot (it isn't clear what is supported, what isn't supported, and what needs a completely different yml entry ... until you try and it fails).

PR #505, opened 2021-09-28
chore(Dockerfile) remove not required packages

locao:


alpine/Dockerfile
```diff
@@ -38,7 +38,7 @@ RUN set -eux; \
     && mv /kong/usr/local/* /usr/local \
     && mv /kong/etc/* /etc \
     && rm -rf /kong \
-    && apk add --no-cache libstdc++ libgcc openssl pcre perl tzdata libcap zip bash zlib zlib-dev git ca-certificates \
+    && apk add --no-cache perl libgcc zlib-dev bash unzip outils-md5 \
```
gszr, line 41: Any thoughts on removing bash, unzip, and outils-md5? These seem unnecessary. Same perhaps for libgcc?

centos/Dockerfile
```diff
@@ -19,13 +19,10 @@ RUN set -ex; \
...
       && echo "$KONG_SHA256  /tmp/kong.rpm" | sha256sum -c -; \
     fi; \
-    yum install -y -q unzip shadow-utils git \
-    && yum clean all -q \
-    && rm -fr /var/cache/yum/* /tmp/yum_save*.yumtx /root/.pki \
     # Please update the centos install docs if the below line is changed so that
     # end users can properly install Kong along with its required dependencies
     # and that our CI does not diverge from our docs.
-    && yum install -y /tmp/kong.rpm \
+    yum install -y /tmp/kong.rpm \
```
gszr, line 25:
```suggestion
    && yum install -y /tmp/kong.rpm \
```

rhel/Dockerfile
```diff
@@ -35,13 +35,10 @@ RUN set -ex; \
...
       && echo "$KONG_SHA256  /tmp/kong.rpm" | sha256sum -c -; \
     fi; \
-    yum install -y -q unzip shadow-utils \
-    && yum clean all -q \
-    && rm -fr /var/cache/yum/* /tmp/yum_save*.yumtx /root/.pki \
     # Please update the rhel install docs if the below line is changed so that
     # end users can properly install Kong along with its required dependencies
     # and that our CI does not diverge from our docs.
-    && yum install -y /tmp/kong.rpm \
+    yum install -y /tmp/kong.rpm \
```
gszr, line 41:
```suggestion
    && yum install -y /tmp/kong.rpm \
```

ubuntu/Dockerfile
```diff
@@ -18,7 +18,8 @@ RUN set -ex \
       && curl -fL https://download.konghq.com/gateway-${KONG_VERSION%%.*}.x-ubuntu-$(cat /etc/os-release | grep UBUNTU_CODENAME | cut -d = -f 2)/pool/all/k/kong/kong_${KONG_VERSION}_$(dpkg --print-architecture).deb -o /tmp/kong.deb \
       && apt-get purge -y curl; \
     fi; \
-    apt-get install -y --no-install-recommends unzip git \
+    # unzip is needed to inject plugins
+    apt install unzip \
```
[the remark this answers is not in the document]
Tieske: Using customize is just one way. Many users have their own scheduled for adding plug-ins. Those will also break if you remove unzip.

The main question we should answer is; does the Kong product require a functional LuaRocks installation?

Maybe that's a question for @PaulFischer-Kong

alpine/Dockerfile
locao: unzip and outils-md5 are needed to add plugins, bash is needed for the smoke tests. We could remove those packages by changing, removing or reducing scope of the smoke tests and documenting that they are needed for adding plugins.

dndx (approved):

gszr:
Hi all.

Let's please not merge this before we get some eyes from Product.

While some of these deps aren't strictly necessary for running Kong, they serve other purposes, as is the case for https://github.com/Kong/docker-kong/tree/master/customize.

cc @PaulFischer-Kong.

alpine/Dockerfile
hutchic: `ca-certificates` has become a somewhat critical package because of let's encrypt. We've had a variety of build break because they either didn't have this at all or had an older version. LE adjusted their root certificates and suddenly the image can't securely talk to a large amount of the internet

Just providing information no thoughts on if it should stay or not

hutchic: cross-linking past conversation regarding `zlib-dev` https://github.com/docker-library/official-images/pull/8337#issuecomment-656859264

gszr:
@locao is this one still required? Can we close it?

locao:
I could not even find the original request for this change. Let's close it, if we want to make this change, it will probably be easier to do it from scratch than rebasing it.

PR #493, opened 2021-08-20
feat(dbless): adjust docker-compose to capture how to run Kong dbless

hutchic:
- requires docker-compose >= 1.28
- need to update docs
- there's no test that dbless mode works

PR #509, opened 2021-10-04
(docfix) minor README change to address latest updates

GGabriele:
Hello! This is just a minor doc fix to address changes in `compose` logic introduced with https://github.com/Kong/docker-kong/pull/493

### DIFF
--- a/compose/README.md
+++ b/compose/README.md
@@ -8,19 +8,9 @@ You can find the official Docker distribution for Kong at [https://hub.docker.co
 
 # How to use this template
 
-This Docker Compose template provisions a Kong container with a Postgres database, plus a nginx load-balancer. After running the template, the `nginx-lb` load-balancer will be the entrypoint to Kong.
-
-To run this template execute:
-
-```shell
-$ docker-compose up
-```
-
-To scale Kong (ie, to three instances) execute:
-
-```shell
-$ docker-compose scale kong=3
-```
+This Docker Compose template supports the following deployments:
+- Kong container with a Postgres database plus a nginx load-balancer, provisioned with `make kong-postgres`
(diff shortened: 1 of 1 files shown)


Tieske:
since this is about docker-compose, we should keep the instructions based on docker-compose, not on the makefile imo. I assume the makefile is mostly used for testing.

wdyt @hutchic ?

GGabriele:
Hi @Tieske, thanks for having a look at this!

In my opinion, the advantage of using a makefile is to abstract certain operations for the users who will only know the specific make targets: next time something change in the `docker-compose.yaml` file (e.g. new flags, new db supported etc), nothing will change on the setup instructions.

Also, `make kong-dbless` clearly states what the end setup would look like, instead of an obscure `docker-compose up -d` , especially if different setups are supported, like in this case.

Obviously, just my 2 cents, up to you! :)

kenmasumitsu:
> # How to use this template
> 
> This Docker Compose template provisions a Kong container with a Postgres database, plus a nginx load-balancer. After running the template, the `nginx-lb` load-balancer will be the entrypoint to Kong.
> 
> To run this template execute:
> 
> ```shell
> $ docker-compose up
> ```
> 
> 

The description is not true. Because Kong run as dbless mode.

I have no option to use `docker compose` or 'make`, could you write down correct description.

If `docker compose` is preferred, 
```
COMPOSE_PROFILES=database KONG_DATABASE=postgres docker-compose up -d
```

or, if `make` is preferred,
```
make kong-postgres
```

kikito (changes requested):
Please change this so that:

* docker-compose is shown first and prioritarily (since it is what will be used in production)
* the makefile is mentioned as a secondary/development aid.

gszr:
Superseded by #628.

PR #515, opened 2021-10-15
feat(alpine) add protobuf-dev dependency

gszr:
It's required by Kong's external plugins using protobuf-based
protocol.

Dependency for other packages is added in https://github.com/Kong/kong-build-tools/pull/409.


aboudreault (approved):

hutchic (approved):

gszr:
Let's please hold on merging this one. Due to differences in how `protobuf-devel` is made available in CentOS 8 (it requires enabling the "PowerTools" repository), I am looking into including the proto files in the Kong packages. The core feature in Kong already supports that, by including `/usr/local/kong/lib` in the list of protobuf include paths.

PR #519, opened 2021-11-01
feat(upgrade): do upgrades when building Kong Enterprise refs: #ENGEN-241

hutchic:
do upgrades only for our enterprise images because we may not do upgrades within our official CE docker images per docker policy


centos/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM centos:8
+FROM centos:7
```
hutchic, line 1: I missed a slack thread where it was decided to not do this downgrade -- reverting

mikefero (approved):
Tested with EE Centos and RHEL 7 releases.

PR #520, opened 2021-11-01
RFC: kong-enterprise release process

hutchic:
This is an RFC to adjust how Kong Enterprise releases are done so at this time review the proposed process not the source code itself.

I'd propose we use docker hub [automated builds ](https://docs.docker.com/docker-hub/builds/) to automatically build Kong Enterprise from github tags. This would effectively make github the single source of truth for Kong CE as well as Kong EE docker images.

Example docker hub build rules
image

source: `/^([0-9]+)\.([0-9]+)\.([0-9]+)\.([0-9]+)$/`
docker tag: `{\1}.{\2},{\1}.{\2}.{\3}.{\4},{\1}.{\2}.{\3}.{\4}-alpine`

reults in
image

when I a 2.6.0.0 tag is pushed to github.

One downside is docker will no longer rebuild when our base images update 😠 

Options are -- monitor ourselves and use the docker hub webhook to rebuild
https://github.com/Al2Klimov/dockerhub-chainreactor
https://hub.docker.com/r/containrrr/watchtower/
https://crazymax.dev/diun/

or we can explicitely pin our base image sha and let dependabot PR updates https://dependabot.com/docker/ and use branches instead of tags for simplicity -- but is also kind of gross


gszr:
Closing this out as stale and superseded by recent build tooling changes.

PR #523, opened 2021-12-03
chore(labels): add some additional static opencontainer spec labels

hutchic:


zekth (dismissed):
:v:

Tieske:
Isn't the `rhel` one missing?

gszr (changes requested):
@hutchic - could you add those to the redhat dockerfile as well?

hutchic:
@gszr redhat has specific container label requirements but I duplicated them over there anyways :+1:

rhel/Dockerfile
```diff
@@ -10,12 +10,17 @@ ENV KONG_VERSION $KONG_VERSION
...
 
 LABEL name="Kong" \
-      vendor="Kong" \
-      version="$KONG_VERSION" \
-      release="1" \
-      url="https://konghq.com" \
-      summary="Next-Generation API Platform for Modern Architectures" \
-      description="Next-Generation API Platform for Modern Architectures"
+    vendor="Kong" \
+    version="$KONG_VERSION" \
```
zekth, line 14: Is it normal that only RHEL has `version` label? Shouldn't it be also propagated to `org.opencontainers.image.version`?

hutchic: the redhat container build process is different than every other container build process. Adding the version label to the other containers is a little more effort than I was up to taking on at this time.

The short version: Kong OSS official docker builds cannot use arg's so we would need to update our automation to sed appropriately

zekth: Fair enough!

Tieske (approved):

PR #545, opened 2022-03-01
Release Kong 2.8.0 and remove CentOS

kikito:
This is the same as #544 , but also removes all mentions of CentOS from everywhere.


hutchic (approved):

wheelerlaw:
For some reason, it is still saying that it is expecting the Centos check, even through it was removed in this PR. Using admin privileges to merge in.

PR #549, opened 2022-03-09
feat(slim): migrate to ubi and debian slim images refs: #FT-2378

hutchic:
additive change to setup slim Dockerfile's. Required some adjustments to tests and customize

related: https://github.com/docker-library/official-images/pull/12093

Follow up PR's to update submit.sh update.sh and make the images slimmer to follow


curiositycasualty:
Doesn't removing the sha mean that we can't have Dependabot update the base image?
Or is the purpose of removing the shas to just let docker handle that funcitonality?

hutchic:
@curiositycasualty yeah with the current way we handle auto updates it does but it seems CI failed when I included them ( https://github.com/docker-library/official-images/runs/5629784259?check_suite_focus=true#step:3:61 )

curiositycasualty:
Waiting on feedback on https://github.com/docker-library/official-images/pull/12093

Dockerfile.deb
```diff
@@ -0,0 +1,47 @@
...
+
+ARG KONG_VERSION=2.8.0
+ENV KONG_VERSION $KONG_VERSION
+
+ARG KONG_SHA256="54a3079ccb9e3751bb096500b25d6d34bdcdf4894e3b34b84b8778a5f99e18d2"
+
+ARG ASSET=remote
+ARG EE_PORTS
+
+COPY kong.deb /tmp/kong.deb
```
flrgh, line 11: I noticed the `kong.rpm` and `kong.apk.tar.gz` assets in this PR are empty (placeholders, I assume) whereas `kong.deb` is not. Any reason for that or maybe an oversight?

hutchic: whoops it should be empty good catch :+1:

curiositycasualty:
FIY: I'm refraining from approving in favor of getting more proper Gateway folk reviews.

Dockerfile.rpm
```diff
@@ -0,0 +1,43 @@
+FROM redhat/ubi8
```
tyler-ball, line 1: Shouldn't this be `ubi8/ubi-minimal` https://catalog.redhat.com/software/containers/ubi8/ubi-minimal/5c359a62bed8bd75a2c3fba8 ? Or is that what you meant by we cannot do official Redhat images?

tyler-ball (commented):
Have you done some test builds based of these? I think the dockerfiles look straightforward, so if a smoke test passed I would feel good about this

Dockerfile.rpm
hutchic: it should not. minimal utilizes a different installation method we don't currently support

customize/install_tools.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+if command -v apk &> /dev/null
+then
+    apk add unzip wget curl
```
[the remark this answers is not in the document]
hutchic: `customize` is the tool customers use to marry their custom plugins into our official docker images https://github.com/Kong/docker-kong/tree/master/customize

These tools don't land on the resulting image as it uses the docker builder pattern

hutchic:
@tyler-ball there's tests within this repository CI as well as within [kong-build-tools ](https://github.com/Kong/kong-build-tools/pull/451)

The kong-build-tools PR demonstrates some of the ways this work is a breaking change FWIW

tyler-ball (approved):
Thanks for answering my questions!

Tieske:
@hutchic I noticed a new `docker-entrypoint.sh` in the root, whilst all images have their own entrypoint script. Is this on purpose, eg. required for future changes?

tests/01-image.test.sh
```diff
@@ -90,45 +101,6 @@ function run_test {
...
-    tfailure
-  else
-    tsuccess
-  fi
-
-
-  # Validate Kong is running as the Kong user (overridden)
-  ttest "Kong is running as the Kong user (overridden)"
-
-  KONG_USER=1001 docker-compose up -d
```
hutchic, line 114: in hindsight I'm unsure why this test was removed <_<

PR #551, opened 2022-03-16
chore(customize): try out a different option

hutchic:
I suspect this will fail but I'm not sure why

### DIFF
--- a/customize/Dockerfile
+++ b/customize/Dockerfile
@@ -26,7 +26,6 @@ RUN /usr/local/openresty/luajit/bin/luajit /packer.lua -- "$INJECTED_PLUGINS" \
 FROM ${KONG_BASE}
 
 COPY --from=build /usr/local /usr/local
-COPY --from=build /docker-entrypoint.sh /docker-entrypoint.sh
 COPY --from=build /old-entrypoint.sh /old-entrypoint.sh
 
 ARG TEMPLATE=empty_file
--- a/customize/Dockerfile
+++ b/customize/Dockerfile
@@ -7,9 +7,6 @@ FROM ${KONG_BASE} AS build
 ARG PLUGINS
 ENV INJECTED_PLUGINS=${PLUGINS}
 
-ARG TEMPLATE=empty_file
-ENV TEMPLATE=${TEMPLATE}
-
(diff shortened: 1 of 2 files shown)


customize/Dockerfile
```diff
@@ -22,16 +19,26 @@ COPY packer.lua /packer.lua
...
 
-RUN /usr/local/openresty/luajit/bin/luajit /packer.lua -- "$INJECTED_PLUGINS"
+RUN /usr/local/openresty/luajit/bin/luajit /packer.lua -- "$INJECTED_PLUGINS" \
+    && /plugins/install_plugins.sh
 
 
 FROM ${KONG_BASE}
 
-COPY --from=build /plugins /plugins
+COPY --from=build /usr/local /usr/local
```
hutchic, line 28: make this less scoped

Tieske:
this can be simplified probably if we copy the entire tree; The packer script works by doing some introspection on the luarocks tree

- it lists all rocks installed
- it lists all plugins installed
- installs the requested plugins
- checks installed plugin+rocks again, and creates a diff-list of stuff that was added (because LuaRocks can pull in more due to dependency resolving)
- *** packs all the rocks that got added
- *** In new image only installs rocks that were added
- In new image adds the plugin names to the KONG_PLUGINS env var

we could probably skip the "***" stages, and instead just copy the entire LuaRocks tree. At least I think that's what your intent was no?

hutchic:
@Tieske more or less.

As part of the work to use slim base images (ubi-slim / debian-slim) I ran into challenges where development tools one would expect to be available weren't. Instead of installing / uninstaling dev tools into the resulting image I thought I'd try to use the existing `build` fat image to do the entire installation then copy the resulting asset into the resulting image.

My lua is rusty but I suspect the steps you mentioned could all be handled by `packer.lua` entirely and not need `install_plugins.sh` if I could get some assistance :pray:

Tieske:
@hutchic I have a new branch for this, but cannot push it now, waiting for my access to this repo to be restored. Closing this for now.

Tieske:
k, seems I cannot close issues without having access either ¯\_(ツ)_/¯

PR #554, opened 2022-04-06
Release: 2.6.1

mashapedeployment:


kikito:
Converting this to draft PR in order to prevent accidental merges.

gszr:
> Converting this to draft PR in order to prevent accidental merges.

good call

tyler-ball:
Do we even need to leave these open? It looks like we only want to merge the `2.8.1` PR if I am reading this correctly

gszr:
> Do we even need to leave these open? It looks like we only want to merge the `2.8.1` PR if I am reading this correctly

Your interpretation is correct. We can safely close it; the tagging can be done on the branch.

kikito:
I have tagged and created releases for both this and 2.7.2. I think that is all that is needed on this project. Closing both PRs.

kikito:
There was one more thing to do: send updated PR to Docker with the new versions and correct tags. Done in https://github.com/docker-library/official-images/pull/12213

PR #564, opened 2022-05-25
chore(slim): make the alpine image slightly more slim

hutchic:
the intention of the more explicit pins is to eventually point dependabot at this repository


curiositycasualty:
Oh I see, you're consolidating the apk add back into one layer and updating to 3.16.0.

curiositycasualty (approved):

PR #566, opened 2022-06-03
feat(slimages): make alpine slimmer by excluding curl from the final image result refs: #ENGEN-633

hutchic:


Dockerfile.apk
```diff
@@ -13,6 +13,7 @@ COPY kong.apk.tar.gz /tmp/kong.apk.tar.gz
 
 RUN set -ex; \
     apk add bash curl ca-certificates; \
+    apk add --no-cache --virtual .build-deps tar gzip; \
```
Tieske, line 16: per description of PR; how would this NOT add curl? it's been added one line up?

hutchic: how very odd ... not sure how I missed that. The curl should be one line lower as `.build-deps` gets removed later in the Dockerfile

hutchic:
Looks like I did the intent of this change properly in https://github.com/Kong/docker-kong/pull/564

PR #577, opened 2022-06-29
chore(non-rhel): unify maintainer to docker@

curiositycasualty:
This uses the new `docker@` alias so that we can more faithfully follow up on feedback re: our containers.
This PR only targets dockerfiles NOT covered by a similar commit on https://github.com/Kong/docker-kong/pull/576 (https://github.com/Kong/docker-kong/pull/576/commits/7e60a40a6995ad71cf6b8996a938cc5aa5b978f3)

### DIFF
--- a/Dockerfile.apk
+++ b/Dockerfile.apk
@@ -1,5 +1,7 @@
 FROM alpine:3.16.0
 
+LABEL maintainer="Kong Docker Maintainers <docker@konghq.com> (@team-gateway-bot)"
+
 ARG KONG_VERSION=2.8.1
 ENV KONG_VERSION $KONG_VERSION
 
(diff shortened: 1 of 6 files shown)


gszr (approved):
LGTM!

gszr (changes requested):
According to our offline conversation, the new label should be named accordingly to https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys.

[This](https://github.com/docker-library/official-images/blob/master/library/kong) should also be updated, but can be deferred to when we release a new kong version (to avoid unnecessary burden on official images maintainers and to not needlessly rebuild our images).

curiositycasualty:
@gszr Let's get this in as a baseline. And then we can expand on all the labels to bring up to OCI compliance in a later PR.

hutchic (approved):

gszr:
Docker maintainers advise against adding non-standard labels. See, e.g., https://github.com/docker-library/official-images/pull/12639. IMO we should update this before the next release cycle, so as to avoid surprises at release day.

PR #579, opened 2022-07-07
fix(rhel): Various Improvements

curiositycasualty:
There's a couple things going on in this PR:
  - explicitly pin `FROM` tags and the kong package URL such that the semi-useless `RHEL_VERSION` can be done away with
  - pinning to an explicit `sha256` also opens the door potentially for dependabot updates
  - ensure that the bash logic causes the build to fail if the package checksum doesn't match; previously the build would continue even if the checksum verification would fail
  - fixup the __currently not run__ `release-rhel` make target

### DIFF
--- a/Dockerfile.apk
+++ b/Dockerfile.apk
@@ -1,5 +1,7 @@
 FROM alpine:3.16.0
 
+LABEL maintainer="Kong Docker Maintainers <docker@konghq.com> (@team-gateway-bot)"
+
 ARG KONG_VERSION=2.8.1
 ENV KONG_VERSION $KONG_VERSION
 
--- a/Dockerfile.deb
+++ b/Dockerfile.deb
@@ -1,5 +1,7 @@
 FROM debian:bullseye-20220509-slim
 
+LABEL maintainer="Kong Docker Maintainers <docker@konghq.com> (@team-gateway-bot)"
+
 ARG KONG_VERSION=2.8.1
 ENV KONG_VERSION $KONG_VERSION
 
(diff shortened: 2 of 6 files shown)


curiositycasualty:
The problem doesn't appear to effect the `ubi8/ubi-minimal` image from the redhat registry, let's use that.

rhel/Dockerfile
```diff
@@ -1,6 +1,4 @@
-ARG RHEL_VERSION=7
-
-FROM registry.access.redhat.com/ubi${RHEL_VERSION}/ubi
+FROM registry.access.redhat.com/ubi7/ubi:7.9@sha256:baccc7f4a9df7f4872116e254efb4e0e1b7d5d8e557947c02d44230ce7b905bc
```
hutchic, line 1: sha256 pins won't work for multi-architecture builds. I think this is currently fine just pointing it out for posterity / future us ;)

rhel-minimal/Dockerfile
```diff
@@ -34,7 +34,8 @@ ENV RHEL_VERSION $RHEL_VERSION
 RUN set -ex; \
     if [ "$ASSET" = "ce" ] ; then \
       curl -fL "https://download.konghq.com/gateway-${KONG_VERSION%%.*}.x-rhel-${RHEL_VERSION}/Packages/k/kong-${KONG_VERSION}.rhel${RHEL_VERSION}.amd64.rpm" -o /tmp/kong.rpm \
-      && echo "$KONG_SHA256  /tmp/kong.rpm" | sha256sum -c -; \
+      && echo "$KONG_SHA256  /tmp/kong.rpm" | sha256sum -c - \
+      || exit 1; \
```
hutchic, line 38: is the `|| exit 1;` necessary?

hutchic (approved):
looks sensible

rhel/Dockerfile
curiositycasualty: Noted!

rhel-minimal/Dockerfile
curiositycasualty: I think so. I was noticing cases where the asset == `ce`, the sha was incorrect, but the build wasn't failing until `kong version` failed to exist.

PR #584, opened 2022-08-02
Add Test for Exposed Ports

curiositycasualty:
As with `EE_PORTS` for enterprise.


curiositycasualty:
Closing until I can get more time to re-think this.

PR #587, opened 2022-08-26
remove the committed docker package and the related Dockerfile COPY

mrwanny:


hutchic (changes requested):
as we transition to 3.0 `alpine/ centos/ rhel/ rhel-minimal/` will all go away. However, this isn't a PR we'd accept as it will break some use cases.

The Dockerfile's in this repository serve a few purposes:
- CE official docker images (which imposes some restrictions on what we can do)
- CE / EE unofficial docker images
- EE official docker images
- Testing Kong assets

When building the CE official image we use the assets from `download.konghq.com`
When building CE unofficial and any EE image we use assets that are from `COPY`
When testing any Kong asset the Kong asset comes from the `COPY`

Capturing this many use cases adds complexity but at the end of the day we know for certain that what we smoke test == unofficial releases == official releases

PR #588, opened 2022-08-26
add  KONG_CLUSTER_CERT and KONG_CLUSTER_CERT_KEY value

mrwanny:
Add support to configure KONG_CLUSTER_CERT and KONG_CLUSTER_CERT_KEY value via environment variables

### DIFF
--- a/alpine/docker-entrypoint.sh
+++ b/alpine/docker-entrypoint.sh
@@ -28,6 +28,18 @@ file_env() {
 
 export KONG_NGINX_DAEMON=${KONG_NGINX_DAEMON:=off}
 
+if [[ -n ${KONG_CLUSTER_CERT_BASE64_VALUE} ]];
+then
+    export KONG_CLUSTER_CERT="/tmp/cluster.crt" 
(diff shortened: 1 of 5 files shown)


hbagdi:
I think we should certainly add support for something like this to Kong - it has been long overdue.
As coincidently, I spent some time yesterday doing this exactly: https://github.com/hbagdi/fly-kong/commit/4f0cf91dfac9a2879efebd20dc6ddaf380df7d2a.

But, this is the wrong place for it.
We should add this into kong/kong itself so that all platforms get it instead of docker itself.

If we were to merge this, it would be the first instance of "docker kong has additional KONG_ that other setups don't" where other setups are non-docker based deployments of Kong. And that goes against Kong's principle of running everywhere and anywhere.

hbagdi:
xref https://github.com/Kong/kong/pull/9253

hutchic (changes requested):
what @hbagdi said :)

Tieske:
related #469

mrwanny:
Closing, looking forward to https://github.com/Kong/docker-kong/pull/469

PR #589, opened 2022-09-02
chore(Dockerfile): adjust where we get the asset into a build-arg

hutchic:
Is it possible to use a `ASSET_LOCATION` build-arg that is defaulted to Kong CE download asset but overridable to a local `kong.deb`?

Will need to check this is suitable with docker folks.
Will need to adjust kong-build-tools accordingly.

### DIFF
--- a/Dockerfile.deb
+++ b/Dockerfile.deb
@@ -3,26 +3,14 @@ FROM debian:bullseye-20220509-slim
 
 LABEL maintainer="Kong Docker Maintainers <docker@konghq.com> (@team-gateway-bot)"
 
-ARG KONG_VERSION=2.8.1
-ENV KONG_VERSION $KONG_VERSION
-
-ARG KONG_SHA256="a57852f49eb6aea6b672a99d1361457b208e8a2d63df3fa362b2d2d6ac979b4c"
-
-ARG ASSET=remote
 ARG EE_PORTS
 
-COPY kong.deb /tmp/kong.deb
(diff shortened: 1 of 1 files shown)


Dockerfile.deb
```diff
@@ -3,26 +3,14 @@ FROM debian:bullseye-20220509-slim
...
 RUN set -ex; \
-    apt-get update; \
-    apt-get install -y curl; \
-    if [ "$ASSET" = "remote" ] ; then \
-      CODENAME=$(cat /etc/os-release | grep VERSION_CODENAME | cut -d = -f 2) \
-      && DOWNLOAD_URL="https://download.konghq.com/gateway-${KONG_VERSION%%.*}.x-debian-${CODENAME}/pool/all/k/kong/kong_${KONG_VERSION}_amd64.deb" \
-      && curl -fL $DOWNLOAD_URL -o /tmp/kong.deb \
-      && echo "$KONG_SHA256  /tmp/kong.deb" | sha256sum -c -; \
-    fi \
-    && apt-get update \
```
hutchic, line 25: I need a conditional sha256 check here but my bash was failing me.

Something along the lines of `if [ ASSET_LOCATION =~ http* ] sha256sum`

gszr:
This is a pretty good refactor idea. I did not know the `ADD` supports remote fetching -- and even a `--checksum` flag. Let's keep this one open and eventually get it ready to review / merge.

PR #613, opened 2022-12-08
chore(apk): remove apk tar gz

mayocream:
KAG-204


chronolaw (approved):

chronolaw:
Does other image has the same problem?

mayocream:
> Does other image has the same problem?

Checked deb and rpm:
https://github.com/Kong/docker-kong/blob/e4d2f21855d6ba3ffd65740691231c1fb670275b/Dockerfile.deb#L31
https://github.com/Kong/docker-kong/blob/e4d2f21855d6ba3ffd65740691231c1fb670275b/Dockerfile.rpm#L52

Only the apk image doesn't clean the tarball.

PR #620, opened 2023-01-12
fix(docker-entrypoint): clean-up dangling unix sockets

locao:
This change adds a clean-up step to remove all sockets that a previous Kong run may have left in $PREFIX directory.

Dangling unix sockets have been a pain point in recent Kong versions and was dealt in traditional Kong in Kong/kong#9254, but the `kong` script is not called when starting Kong inside a container, so a similar change to `docker-entrypoint.sh` was added here. 

FTI-4525

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -45,9 +45,18 @@ if [[ "$1" == "kong" ]]; then
     kong prepare -p "$PREFIX" "$@"
 
     # remove all dangling sockets in $PREFIX dir before starting Kong
+    LOGGED_SOCKET_WARNING=0
     for localfile in "$PREFIX"/*; do
-      if [ -S $localfile ]; then
-        rm -f $localfile
+      if [ -S "$localfile" ]; then
+        if (( LOGGED_SOCKET_WARNING == 0 )); then
+          printf >&2 'WARN: found dangling unix sockets in the prefix directory '
+          printf >&2 '(%q) ' "$PREFIX"
+          printf >&2 'while preparing to start Kong. This may be a sign that Kong '
+          printf >&2 'was previously shut down uncleanly or is in an unknown state '
+          printf >&2 'and could require further investigation.\n'
+          LOGGED_SOCKET_WARNING=1
+        fi
+        rm -f "$localfile"
       fi
     done
 
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -44,6 +44,13 @@ if [[ "$1" == "kong" ]]; then
   if [[ "$2" == "docker-start" ]]; then
     kong prepare -p "$PREFIX" "$@"
 
+    # remove all dangling sockets in $PREFIX dir before starting Kong
+    for localfile in "$PREFIX"/*; do
+      if [ -S $localfile ]; then
+        rm -f $localfile
+      fi
+    done
+
     ln -sf /dev/stdout $PREFIX/logs/access.log
     ln -sf /dev/stdout $PREFIX/logs/admin_access.log
     ln -sf /dev/stderr $PREFIX/logs/error.log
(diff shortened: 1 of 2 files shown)


docker-entrypoint.sh
```diff
@@ -44,6 +44,13 @@ if [[ "$1" == "kong" ]]; then
   if [[ "$2" == "docker-start" ]]; then
     kong prepare -p "$PREFIX" "$@"
 
+    # remove all dangling sockets in $PREFIX dir before starting Kong
+    for localfile in "$PREFIX"/*; do
+      if [ -S $localfile ]; then
```
flrgh, line 49: Need to quote `$localfile` or use `[[` in this case:

```suggestion
      if [[ -S $localfile ]]; then
```

docker-entrypoint.sh
```diff
@@ -44,6 +44,13 @@ if [[ "$1" == "kong" ]]; then
...
+        rm -f $localfile
```
flrgh, line 50:
```suggestion
        rm -f "$localfile"
```

docker-entrypoint.sh
```diff
@@ -44,6 +44,13 @@ if [[ "$1" == "kong" ]]; then
   if [[ "$2" == "docker-start" ]]; then
     kong prepare -p "$PREFIX" "$@"
 
+    # remove all dangling sockets in $PREFIX dir before starting Kong
```
flrgh, line 47: Having leftover sockets is not something that happens under normal circumstances, so I think we should log a warning here.

Something like...

```sh
LOGGED_SOCKET_WARNING=0

if [[ -S $localfile ]]; then
  if (( LOGGED_SOCKET_WARNING == 0 )); then
    printf >&2 'WARN: found dangling unix sockets in the prefix directory '
    printf >&2 '(%q) ' "$PREFIX"
    printf >&2 'while preparing to start Kong. This may be a sign that Kong '
    printf >&2 'was previously shut down uncleanly or is in an unknown state '
    printf >&2 'and could require further investigation.\n'

    LOGGED_SOCKET_WARNING=1
  fi

  # 1. using the verbose (-v) flag
  # 2. explicitly sending output to stderr
  rm -fv "$localfile" >&2
fi
```

The warning here is copy/pasted from [kong start](https://github.com/Kong/kong/blob/d7e3cbc7ac2f28fa0cc057859203b01e61f993f7/kong/cmd/start.lua#L30-L34)

flrgh:
There's a nearly identical entrypoint script at [ubuntu/docker-entrypoint.sh](https://github.com/Kong/docker-kong/blob/a3105a6f2046fc46af694e941c32fbb32e2b0389/ubuntu/docker-entrypoint.sh). Do these changes need to be added there as well?

flrgh (changes requested):
Mostly good to go! I requested a couple changes for bash quirks and some extra verbosity.

docker-entrypoint.sh
```diff
@@ -44,6 +44,13 @@ if [[ "$1" == "kong" ]]; then
...
+    for localfile in "$PREFIX"/*; do
```
flrgh, line 48: I doubt it would ever come up (certainly not with any sockets that Kong currently creates), but worth noting that this will miss files that start with `.` (i.e. `$PREFIX/.my_socket.sock`) depending on the value of `dotglob` ([docs](https://www.gnu.org/software/bash/manual/html_node/The-Shopt-Builtin.html) since they are supposed to be "hidden" by default.

...bash is so finicky and weird.

Tieske:
> There's a nearly identical entrypoint script at [ubuntu/docker-entrypoint.sh](https://github.com/Kong/docker-kong/blob/a3105a6f2046fc46af694e941c32fbb32e2b0389/ubuntu/docker-entrypoint.sh). Do these changes need to be added there as well?

Yes, they should.

docker-entrypoint.sh
locao: Good idea

locao:
> There's a nearly identical entrypoint script at [ubuntu/docker-entrypoint.sh](https://github.com/Kong/docker-kong/blob/a3105a6f2046fc46af694e941c32fbb32e2b0389/ubuntu/docker-entrypoint.sh). Do these changes need to be added there as well?

Good catch! Added the same set of changes to that file.

flrgh (approved):

PR #632, opened 2023-03-08
WIP

Tieske:


Tieske:
@gszr I updated the old PR, can you have a look?

Dockerfile.apk
```diff
@@ -30,11 +30,14 @@ RUN set -ex; \
     fi \
     && tar -C / -xzf /tmp/kong.apk.tar.gz \
     && apk add --no-cache libstdc++ libgcc perl tzdata libcap zlib zlib-dev bash \
-    && adduser -S kong \
     && addgroup -S kong \
+    && adduser -S -G kong kong \
     && mkdir -p "${KONG_PREFIX}" \
-    && chown -R kong:0 ${KONG_PREFIX} \
-    && chown kong:0 /usr/local/bin/kong \
+    && chown -R root:kong ${KONG_PREFIX} \
```
gszr, line 36: @Tieske Is there more context on the motivation for this change? IIRC, the owner of kong files was changed to `kong` fairly recently, so I am afraid of changing it back to root.

Tieske: In the original pr by Colin; Changing the ownership to root but giving the Kong user access prevents the Kong user from modifying the files.

fffonion:
file permission of deb and rpm should be set by `postinstall` script so that users install binary packages directory
should have the same behaviour of docker images.

Dockerfile.apk
fffonion: generally, everything under a common shared path like `/usr/local/lib`, `/usr/local/share` should always be owned by `root:root` because technically other packages will also write to it. this applies to also /etc/kong/kong.conf; only
/usr/local/kong (and /usr/local/openresty, if we mark our package a conflict with `openresty`) should be owned by
`kong:kong`.

Tieske:
@fffonion not sure what you're saying. What needs to change about this PR? all should be `root:root`?

Dockerfile.apk
gszr: The course of action is IMO:
- Consolidate what the permissions should be (I tend to agree with most of what @fffonion said)
- Move permission-related code to post-install scripts (alpine is an exception since it's not really a distro package) as suggested by Wangchong. I cannot remember of any particular issues around this explaining why they were placed in the docker images only.

Tieske:
ping @gszr @fffonion

fffonion:
@Tieske sry was typing on phone. i was mentioning to move the `chown`s to here https://github.com/Kong/kong/blob/master/build/package/postinstall.sh, for deb and rpm; i'm not sure if alpine packages take postinstall scripts, if they don't, leave it here at docker-kong is fine. so that we don't end up having two different states of files ownership when user are installing binary packages directly, vs using docker.

and beside of that, me and @gszr both agree that most files we shipped should be owned by `root:root`, with the exception of `/usr/local/kong` directory.

hbagdi:
ping  @Tieske @gszr

gszr:
> and beside of that, me and @gszr both agree that most files we shipped should be owned by `root:root`, with the exception of `/usr/local/kong` directory. the runtime directory can be owned by `kong:kong` or `kong:root` whichever fits our needs.

Could this break custom plugin installs attempted with the `kong` user?

PR #634, opened 2023-03-09
fix(*): use `ln -sf` to overwrite symlink if it exists

hanshuebner:
Versions of Kong built with Bazel include the symbolic links to /usr/local/bin which made the Dockerfiles fail.

KAG-817


Tieske (approved):
lgtm (in anticipation of green CI)

PR #641, opened 2023-03-29
Fix 3.2 changes

AntoineJac:
Fix for 3.2 changes:
- file seems to be called entrypoint.sh no more docker-entrypoint
- missing symlinks https://kongstrong.slack.com/archives/C0496KBU16E/p1676401319527349
https://konghq.atlassian.net/browse/KAG-617


Tieske:
interesting; why did the tests on `master` not fail?

PR #688, opened 2023-12-01
Docker Compose V2 support

Mizunanari:
## Reason for Change

- "version" is no longer necessary
- `docker-compose` has been replaced by `docker compose`.

## Reference Links

- https://matsuand.github.io/docs.docker.jp.onthefly/compose/compose-file/#version-top-level-element
- https://matsuand.github.io/docs.docker.jp.onthefly/compose/cli-command/


Mizunanari:
## Reason for Change

- docker-compose.yaml changed to compose.yaml. standard file name changed in V2.

## Reference Links

https://docs.docker.com/compose/environment-variables/envvars/#compose_file

PR #739, opened 2026-05-13
Fix docker compose

thesp0nge:
In order to have the kong user as regular one and not a DBA, a custome init script it was created and the docker compose files changed accordingly

### DIFF
--- a/compose/init_kong_db_user.sql
+++ b/compose/init_kong_db_user.sql
@@ -1,0 +1,16 @@
+\getenv password_file_path KONG_PASSWORD_FILE
+
+\if :{?password_file_path}
+  \set kong_password `cat :password_file_path`
+\else
+  \set kong_password 'kong'
+\endif
+
+CREATE USER kong WITH NOSUPERUSER NOCREATEDB NOCREATEROLE PASSWORD :'kong_password';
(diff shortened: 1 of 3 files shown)


thesp0nge:
Do we have any update on this?

compose/init_kong_db_user.sql
```diff
@@ -0,0 +1,16 @@
+\getenv password_file_path KONG_PASSWORD_FILE
+
+\if :{?password_file_path}
+  \set kong_password `cat :password_file_path`
+\else
+  \set kong_password 'kong'
+\endif
+
+CREATE USER kong WITH NOSUPERUSER NOCREATEDB NOCREATEROLE PASSWORD :'kong_password';
+CREATE DATABASE kong OWNER kong;
```
locao, line 10: Before your changes, db and user were configurable (`POSTGRES_DB: ${KONG_PG_DATABASE:-kong}` and `POSTGRES_USER: ${KONG_PG_USER:-kong}`) but now both will be forced to be `kong`. Is it intentional? Could we keep the previous behavior?

thesp0nge: This is because I create kong db user as a non DBA

